PR #154, opened 2014-10-17
allow to use additional sources defined in spec file

misli:
Hi,
sometimes you may need to use additional source (other than the one generated from git).
This patch checks for sources specified in spec file and gets missing ones from current directory.
It might be better to use some commandline option to specify the location of additional sources.
Take it just like a food for thought.


dgoodwin:
Hi @misil, any chance you could add a unit test to demonstrate usage, and I can make sure we keep this working going forward?

PR #157, opened 2014-11-24
Python3

xsuchy:
Use python3 on Fedora22+
and do not compress man pages (because rpmbuild do that automatically).


xsuchy:
ping! Any progress here?

dgoodwin:
Looks good, thanks.

PR #162, opened 2015-02-28
try to find spec file in some subdirectories too

xsuchy:
Some projects hesitate to put spec file in top levele directories and use "packaging" or "dist". Try to look there too.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -229,20 +229,31 @@ def create_builder(package_name, build_tag,
     return builder
 
 
-def find_file_with_extension(in_dir=None, suffix=None):
+def find_file_with_extension(in_dir=None, suffix=None, top=True):
     """ Find the file with given extension in the current directory. """
     if in_dir is None:
         in_dir = os.getcwd()
     file_name = None
     debug("Looking for %s in %s" % (suffix, in_dir))
+    if not os.path.isdir(in_dir):
+        return None
     for f in os.listdir(in_dir):
         if f.endswith(suffix):
             if file_name is not None:
                 error_out("At least two %s files in directory: %s and %s" % (suffix, file_name, f))
             file_name = f
-            debug("Using file: %s" % f)
+            debug("Using file: %s" % os.path.join(in_dir, f))
     if file_name is None:
-        error_out("Unable to locate a %s file in %s" % (suffix, in_dir))
+        if top:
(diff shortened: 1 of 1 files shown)


xsuchy:
ping?

src/tito/common.py
```diff
@@ -229,20 +229,31 @@ def create_builder(package_name, build_tag,
...
         if f.endswith(suffix):
             if file_name is not None:
                 error_out("At least two %s files in directory: %s and %s" % (suffix, file_name, f))
             file_name = f
-            debug("Using file: %s" % f)
+            debug("Using file: %s" % os.path.join(in_dir, f))
     if file_name is None:
-        error_out("Unable to locate a %s file in %s" % (suffix, in_dir))
+        if top:
+            sub_dir_list = ["distribution", "dist", "packaging"]
```
jumanjiman: i also see projects put the spec file in `contrib` and `spec` subdirs

jumanjiman: perhaps there should be a tito property to specify the subdir in which spec file lives?

dgoodwin: This alone is probably a pretty good sign we shouldn't try to guess the location, too many possibilities.

For a single project repo, dropping into tito.props seems like a good idea. That won't work so hot for multi-project repos though. Perhaps we define in tito.props, and assume to look in that sub-dir relative to the package dir.

dgoodwin:
Apologies for the delay here but this really needs unit tests, otherwise I'm going to end up breaking it or merging something that does and then not realize. 

Could you add some unit tests to verify it's working for both singleproject and multi-project tito repos? Should be good examples to start with in test/functional/singleproject_tests.py and multiproject_tests.py.

src/tito/common.py
dgoodwin: Also would searching for a spec file be reasonable?

jumanjiman: let's see we search and find a `.spec` file.
we then need to figure out:

* is it really an rpm spec file?
* where does the spec file live in relation to the top of its dir tree, especially wrt multi-project repo

xsuchy: > Also would searching for a spec file be reasonable?

Rather not. That could be just example. Test file... whatever. However I agree with configuration variable in tito.props.

dgoodwin:
Going to close for now, please reopen when ready if this is to be pursued.

PR #163, opened 2015-03-14
Add forceversiontagger buildconfig for use with --use-version

sdodson:


dgoodwin:
Apologies I think I'm going to tentatively nack this, it's making a convoluted situation a little more convoluted. 

There's a TODO on ForceVersionTagger to incorporate it into the main tagger, which conceptually makes a lot of sense, we're just handing in the value to set for version/release instead of letting it do it itself. It should be possible. 

Not super keen on defining two taggers in the config file, one for with --use-version and one without. I'd say lets try to get --use-version passing it's data into whatever tagger is configured. Then you can just have your custom tagger defined once.

PR #170, opened 2015-04-23
Allow override of rpmbuild_options from builder arguments

### LINKED ISSUE #155: Is not possible to provide rpmbuild-options when releasing a package
tito build --rpmbuild-options='--nodeps' work flawless but when I try to release a package with this rpmbuild-option the parameter never reaches the builder. 
I've tried with the --arg option, but this provides custom argument to the builder, not to rpmbuild.

domcleal:
Instead of being a command line-only option, the string of extra rpmbuild
arguments can now be passed in via the general builder args mechanism, so it
can even be configured from releasers.conf.

Fixes #155


dgoodwin:
Tests clear, merged, thanks!

PR #172, opened 2015-04-24
Fix tarball timestamps from git archive with Python.

awood:
Tito passes "git archive" a tree ID.  The "git archive" man page states:

    git archive behaves differently when given a tree ID versus when
    given a commit ID or tag ID. In the first case the current time
    is used as the modification time of each file in the archive.

Using the current time means that every time we build the source
tarball, the file fingerprint will change since the metadata in the
tarball changes.  We don't want that since build systems track the
fingerprint to see if the actual source has changed.

This process was previously handled in an enigmatic Perl script that
lacked any comments whatsoever.  Converting it to well-commented Python
makes the process less mysterious and speedier since Tito doesn't need
to shell out to Perl.


src/tito/tar.py
```diff
@@ -0,0 +1,296 @@
...
+        else:
+            self.process_header(chunk_props)
+            self.process_file_data(chunk_props['size'])
+
+    def fix(self):
+        try:
+            chunk = self.fh.read(RECORD_SIZE)
+            while chunk != "" and not self.done:
+                self.process_chunk(chunk)
+                chunk = self.fh.read(RECORD_SIZE)
```
xsuchy: This is not correct. read() will read *at most* RECORD_SIZE, but it can read only part of it - and still not be at EOF. While this is unusual, it can happen. You should put there the second while-loop as in original perl script.

awood: > This is not correct. read() will read at most RECORD_SIZE, but it can read only part of it - and still not be at EOF. While this is unusual, it can happen. You should put there the second while-loop as in original perl script.

Good point.  Addressed in dc47b4c.

dgoodwin:
Looks like a lot of errors on every platform in the docker tests, not sure what's going on there but they look pretty legit. Check out hacking/runtests.sh

/tmp/titotest-centos-6-python.out       : FAILED (errors=1)
/tmp/titotest-fedora-21-python.out      : FAILED (errors=1)
/tmp/titotest-fedora-rawhide-python.out : FAILED (errors=1)
/tmp/titotest-fedora-21-python3.out     : FAILED (errors=14)
/tmp/titotest-fedora-rawhide-python3.out: FAILED (errors=14)

dgoodwin:
Make sure to merge master first, docker test suite just recently got running again.

awood:
I will take a look.  Since this code actually became relevant in the MEAD stuff, I've rebased the MEAD branch onto this.  I'll probably close this PR.

Issue #161, opened 2015-02-28
substitue /rel-eng for /.tito


xsuchy:
I find that some projects hesitate to use Tito, because it creates that /rel-eng directory and they take it as unwanted pollution. Yet they are able to accept .tx/, .travis/ etc.

We may get more users if we create /.tito instead of /rel-eng directory.

Opinions? Comments?
I am willing to create backward compatible patch if I get enough +1.

jumanjiman:
:+1:

I like the association with Tito.

I have to point out that .travis.yml and so forth are "extras". One can build project without Travis. Once a project uses Tito, it somewhat forces a workflow. It's a comfortable workflow, but the tie-in is obvious.

domcleal:
:+1:

dgoodwin:
bd (that's a double thumbs up)

ktdreyer:
I always wondered why there was a `rel-eng` directory. I figured it was because tito came from deep inside of Red Hat's rel-eng team?

xsuchy:
@ktdreyer Nope. It was born in Spacewalk git (https://github.com/spacewalkproject/spacewalk) and there have been dozens packages and we kept them in hierarchical and everything release engineer oriented went into /rel-eng/ directory. Even files which were later separated into project which got name Tito. And nobody changed that name, because there was no reason. End of history lesson :)

PR #173, opened 2015-04-26
substitute /releng for /.tito #161

xsuchy:
if /releng is present it will be used. Otherwise /.tito will be used.

The patch is surprisingly small. With major changes in documentation.
I decided to not declare /rel-eng/ as obsolete and ask people for rename as I would like to see this function for at least one release before we call /rel-eng/ obsolete (if ever).

### DIFF
--- a/README.mkd
+++ b/README.mkd
@@ -52,7 +52,7 @@ From your git repository:
 
     tito init
 
-This will create a top-level metadata directory called "rel-eng/" and commit it
+This will create a top-level metadata directory called ".tito/" and commit it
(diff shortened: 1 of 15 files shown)


jumanjiman:
:+1: 

after this commit:

* new projects get initialized with `.tito` dir
* existing projects continue to use `rel-eng` dir
* convert existing project to new dir is as simple as `git mv rel-eng .tito`

xsuchy:
*nod*

dgoodwin:
Very nice, thanks,

PR #176, opened 2015-05-06
bash-completion for tito command

jflorian:
Here's an initial go at getting some bash-completion for the tito command.
Module options are a union of those culled from both "tito MODULE --help" and
"man tito" -- there are discrepancies.  Some options were deliberately omitted
such as those documented as deprecated or "avoid using this please".  All
short options are also omitted as their completion would not save any typing
but would hinder the accessibility of the longer options.

### DIFF
--- a/share/tito_completion.sh
+++ b/share/tito_completion.sh
@@ -1,0 +1,123 @@
+# Copyright (c) 2015 John Florian <jflorian@doubledog.org>
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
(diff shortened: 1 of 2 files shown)


dgoodwin:
Cannot accept as this discriminates against zsh users. 

llama

dgoodwin:
Kidding of course. But seriously, someone else will have to tell me if this stops working.

jflorian:
On 05/14/2015 02:44 PM, Devan Goodwin wrote:
>
> Cannot accept as this discriminates against zsh users.
>
> llama
> <https://cloud.githubusercontent.com/assets/51265/7639103/036d9e10-fa50-11e4-9888-394ae76e488f.gif>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dgoodwin/tito/pull/176#issuecomment-102132305>.
>

That llama gives me the willies.  I thought only border collies had an
evil eye!

I've long wanted to switch to zsh but am required to work on many a host
where it's not available and as a creature of habit I don't do
split-brain very well.

I thought zsh had a way to use bash-completion scripts.  Or is that just
my imagination?

PR #177, opened 2015-05-15
Add Maven support to Tito

awood:
This PR adds support for the "maven-build" command in `rhpkg`.

Please catch me in IRC for details around testing this.

***Make sure you are running a newish version of `rhpkg`.  I'm on 1.22.  Older versions had a fatal bug in them that kept `maven-build` from working.***


src/tito/release/distgit.py
```diff
@@ -387,6 +395,98 @@ class DistGitReleaser(FedoraGitReleaser):
...
+        self.mead_url = "%s#%s" % (
+            self.mead_scm,
+            self.builder.build_tag)
+        self.push_url = self.releaser_config.get(self.target, "mead_push_url")
+
+    def _sync_mead_scm(self):
+        with chdir(self.git_root):
+            print("Syncing local repo with %s" % self.push_url)
+            try:
+                run_command("git push %s %s" % (self.push_url, self.builder.build_tag))
```
dgoodwin: Just the one tag being built? Nice.

dgoodwin:
Looks like a problem with cloning the repo during sync having to do with crossing devices locally:

Got package metadata: ['0.9.50-1', 'server/']
Working in: /tmp/tito/release-candlepinTASodl
Syncing local repo with git+ssh://git.engineering.redhat.com/srv/git/users/awood/candlepin.git

########## ERROR ############
Error running command: git clone --local /home/dgoodwin/src/candlepin /tmp/tito/maven-clone-candlepinAbfV5o
Status code: 32768
Command output: Cloning into '/tmp/tito/maven-clone-candlepinAbfV5o'...
fatal: failed to create link '/tmp/tito/maven-clone-candlepinAbfV5o/.git/objects/info/packs': Invalid cross-device link
fatal: The remote end hung up unexpectedly

Probably just need to drop --local and do a literal copy? (or possibly --no-hardlinks?)

src/tito/release/distgit.py
```diff
@@ -387,6 +395,98 @@ class DistGitReleaser(FedoraGitReleaser):
...
+            except RunCommandException as e:
+                if "rejected" in e.output:
+                    if self._ask_yes_no("The remote rejected a push.  Force push? [y/n] ", False):
+                        run_command("git push --force %s %s" % (self.mead_scm, self.builder.build_tag))
```
dgoodwin: Hmm what's the differentiation between mead_scm and mead_push_url, and why do we switch from one to the other here when the user specified a force push?

Should mead_push_url be marked as a required param as well?

We should get these and the whole builder documented in doc/builders.mkd.

dgoodwin: Looks like mead_push_url may actually be needed, or at least tito errors if not present.

dgoodwin:
I don't think release --scratch it making it's way to rhpkg maven-build, this might be handy and worth hooking up.

I think dry-run may be pushing the reference to the mirror repo, is this fairly dangerous? Should we just error out if you try? Or is it possible to get true a dry-run with this?

dgoodwin:
Overall this is fantastic, firstly that you managed to keep local builds working despite the templates, and secondly for the ramifications for Candlepin. 

687474703a2f2f692e696d6775722e636f6d2f477a6837314b752e676966

src/tito/release/distgit.py
awood: > Hmm what's the differentiation between mead_scm and mead_push_url, and why do we switch from one to the other here when the user specified a force push?

These two parameters are a little wart that exist due to the implementation in `rhpkg`.  `rhpkg` has certain URLs whitelisted as acceptable places to build from.  In our case, the 'srv/git/' in the URL we have to use to push to flunks the whitelist check.  There didn't seem to be much interest in fixing/changing that on the `rhpkg` side of the house.

> Should mead_push_url be marked as a required param as well?

Yes.  I will fix.

awood:
> I think dry-run may be pushing the reference to the mirror repo, is this fairly dangerous? Should we just error out if you try? Or is it possible to get true a dry-run with this?

Will fix.

dgoodwin:
I see that a username is required for the repository access, which means we can't really specify our own usernames as releasers.conf is under source control. 

Should we perhaps implement a MEAD_SCM_USERNAME in ~/.titorc and substitute it in automatically? (and error out if it's not set) See man 5 titorc.

dgoodwin:
Did you have ideas for functional testing? Because I can't quite see how it's possible with that scm step in the middle...

awood:
> Did you have ideas for functional testing? Because I can't quite see how it's possible with that scm step in the middle...

I haven't really thought of anything unfortunately.  How does Tito do functional testing today for other dist-git releasers?

jmrodri:
:-1: :cry: :sob: :no_entry_sign: :no_good:

jmrodri:
no

jmrodri:
no

awood:
@jmrodri We're still using Buildr to do everything.  In fact Buildr generates the pom files for us. :gem: :dollar:

dgoodwin:
I'm afraid we don't do any functional testing of dist git style releasers. :(

jluaquo

Just too complicated with all the auth/credentials that would be required and such. Not even something easy to mock really. :(

dgoodwin:
And any maven haters need to attend the demo tomorrow to see exactly what this means. :)

what

dgoodwin:
Colorized output?

ux7vbuv

Lots of goodies in here. Checksums look to be the same before/after. I haven't tried running the MEAD builds myself but we'll deal with whatever comes up later. This is great stuff, thanks!

PR #179, opened 2015-06-08
try to find spec file in some subdirectories too

xsuchy:
Some projects hesitate to put spec file in top levele directories and use "packaging" or "dist". Try to look there too.

This is second take after #162. This time with test.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -229,20 +229,31 @@ def create_builder(package_name, build_tag,
     return builder
 
 
-def find_file_with_extension(in_dir=None, suffix=None):
+def find_file_with_extension(in_dir=None, suffix=None, top=True):
     """ Find the file with given extension in the current directory. """
     if in_dir is None:
         in_dir = os.getcwd()
     file_name = None
     debug("Looking for %s in %s" % (suffix, in_dir))
+    if not os.path.isdir(in_dir):
+        return None
     for f in os.listdir(in_dir):
         if f.endswith(suffix):
             if file_name is not None:
                 error_out("At least two %s files in directory: %s and %s" % (suffix, file_name, f))
             file_name = f
-            debug("Using file: %s" % f)
+            debug("Using file: %s" % os.path.join(in_dir, f))
     if file_name is None:
-        error_out("Unable to locate a %s file in %s" % (suffix, in_dir))
+        if top:
+            sub_dir_list = ["distribution", "dist", "packaging", "spec", "contrib"]
+            for sub_dir in sub_dir_list:
+                file_name = find_file_with_extension(os.path.join(in_dir, sub_dir), suffix, top=False)
+                if file_name:
+                    return os.path.join(sub_dir, file_name)
+            error_out("Unable to locate a {0} file in {1} (and in {2} sub directories)".format(suffix,
+                in_dir, ', '.join(sub_dir_list)))
+        else:
+            return None
     else:
         return file_name
 
--- a/test/functional/fixture.py
+++ b/test/functional/fixture.py
@@ -188,3 +188,16 @@ class TitoGitTestFixture(unittest.TestCase):
 
         if tag:
             tito('tag --keep-version --debug --accept-auto-changelog')
+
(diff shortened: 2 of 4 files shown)


dgoodwin:
Still a ways to go with this I'm afraid. I'm much more worried about tito actually working than find file working, we need to test that tito actually runs correctly with this.

Today, when you run tito, we error out fast if you run it in the wrong dir, i.e. not the project root. With this patch you can now mistakenly run tito in lets say "packaging", and end up with a broken tag/package. (note the relative path that lands in .tito/packages/pkgname) Here's a multi-project test to verify what I'd like to see:

    def test_spec_subdir_tag_in_wrong_dir(self):
        # Should fail if user mistakenly runs this in wrong directory:
        os.chdir(os.path.join(self.repo_dir, 'pkg3', 'packaging'))
        self.assertRaises(SystemExit, tito, 'tag --debug --accept-auto-changelog')

I'd also like to see tests for tag and build passing, both for multi and single project, something similar to this:

    def test_spec_subdir_tag_and_build(self):
        os.chdir(os.path.join(self.repo_dir, 'pkg3'))
        tito('tag --debug --accept-auto-changelog')
        with open (os.path.join(self.repo_dir, ".tito/packages/", TEST_PKG_3), "r") as myfile:
            pkg_metadata = myfile.read().replace('\n', '')
            self.assertEquals("0.0.2-1 pkg3/", pkg_metadata)

        os.chdir(os.path.join(self.repo_dir, 'pkg3'))
        artifacts = tito('build --rpm')
        self.assertEquals(3, len(artifacts))

Note however this is failing, the spec file once exported is landing in both the top level of the tree and the sub-dir, and the build is looking for 0.0.1, not sure why this is, probably a bug in the test but I don't quite have time to trace it down right now.

src/tito/common.py
```diff
@@ -229,20 +229,31 @@ def create_builder(package_name, build_tag,
...
         if f.endswith(suffix):
             if file_name is not None:
                 error_out("At least two %s files in directory: %s and %s" % (suffix, file_name, f))
             file_name = f
-            debug("Using file: %s" % f)
+            debug("Using file: %s" % os.path.join(in_dir, f))
     if file_name is None:
-        error_out("Unable to locate a %s file in %s" % (suffix, in_dir))
+        if top:
+            sub_dir_list = ["distribution", "dist", "packaging", "spec", "contrib"]
```
dgoodwin: Can we avoid guessing and hardcoding a list here, and instead make it a config property in tito.props? Assume that it's single valued, and if it's a multi-project repo, we expect to find the spec file there for *all* projects in the repo.

src/tito/common.py
```diff
@@ -229,20 +229,31 @@ def create_builder(package_name, build_tag,
     return builder
 
 
-def find_file_with_extension(in_dir=None, suffix=None):
+def find_file_with_extension(in_dir=None, suffix=None, top=True):
```
dgoodwin: Unless I'm missing something I don't think the "top" param is actually used anywhere.

xsuchy: It is used on lines 247 and 250.

ashcrow:
I may pick up on this. I'm looking to use a different location than root as well.

Conan-Kudo:
@ashcrow @xsuchy @dgoodwin Rather than attempting to find it, why not just let it get defined as a property so tito knows where it is, as I described in #306?

dgoodwin:
@xsuchy I've got some comments in here that don't read so nicely in retrospect, my apologies.

Defining as a property in tito.props sounds ok to me, multi-project repos would be the challenge but we could assume it's the same for all projects. Just a little tricky because the location of that file has been considered the main way we identify where a sub-project exists, so there might be assumptions baked in unusual places.

Conan-Kudo:
@dgoodwin If you don't have per-project props, then yeah, assuming it's the same for all would work.

dgoodwin:
IIRC we do have project props but it's at the same top dir of the project location, and if you don't want a spec there you probably don't want a tito.props there, and now things are getting confusing. :)

PR #180, opened 2015-06-11
Add RHPKG/FEDPKG_USER to be passed to rh/fedpkg

eLobato:
Specifying these values on .titorc will allow to use tito without having to create a specific username in Linux for it, in case your username in git is different than the one on your computer.

### DIFF
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -23,6 +23,7 @@ from tito.compat import getoutput, getstatusoutput, write
 from tito.release import Releaser
 from tito.release.main import PROTECTED_BUILD_SYS_FILES
 from tito.buildparser import BuildTargetParser
+from tito.cli import read_user_config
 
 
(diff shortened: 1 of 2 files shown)


jmrodri:
I like this idea, I'm not sure how I got around this before. I have the vague recollection that I would set an ENVIRONMENT variable before running tito. But I can't find an example of that in my history.

:+1: Needs to be tested by someone before merging. If I have time I'll give it a run.

eLobato:
@jmrodri Maybe it's the RSYNC_USERNAME env variable that made you think that :smile: ! I asked @dgoodwin and he preferred the .titorc route.

Thanks for testing!

jmrodri:
@eLobato HAH! It was RSYNC_USERNAME! :smile: thanks.

dgoodwin:
This looks great, thank you, going to merge and make a couple adjustments to avoid the double user config read. Will be in next release.

PR #181, opened 2015-06-12
Filter lines beginning with "Merge" from the changelog.

awood:
Merge commits rarely add value to the RPM changelog.  This patch removes
lines from the changelog that begin with "Merge pull request" or "Merge
remote-tracking" or "Merge branch".  The old behavior can be restored
by setting "keep_merge_commits" to 1.


domcleal:
I think you can also use `git log --no-merges`, which might be more accurate than a regexp?

awood:
> I think you can also use git log --no-merges, which might be more accurate than a regexp?

Good suggestion.  I've done it that way instead.  Thanks!

PR #183, opened 2015-06-15
Provide ability to turn off colored output.  Fixes #182.

### LINKED ISSUE #182: Coloured output needs to be configurable
d489dd9 / 53144dc added coloured output to tito, but the blue is rather hard to read for me on a black background.

screenshot from 2015-06-15 12 55 46

tito should probably:

1. have a simple on/off config option for coloured output
2. have at least two built in colour schemes (dark/light)
3. possibly add the ability to configure colour schemes

awood:


domcleal:
Thanks @awood!

PR #189, opened 2015-07-24
fixes(188): Run builder cleanup if SCL for KojiReleaser

ehelms:


src/tito/release/main.py
```diff
@@ -527,6 +527,7 @@ def _koji_release(self):
                 if scl:
                     builder = copy.copy(self.builder)
                     builder.scl = scl
+                    builder.cleanup()
```
domcleal: I'm unsure if calling cleanup could leave the *original* builder (self.builder) in a strange state if it has already been used?  (The source dir etc might have been cleaned, but self.builder thinks it's dirty still.)

ehelms: Sorta ad-hoc testing changes to start with, this worked for me originally for scl and non-scl, but @jlsherril is reporting failures after adding the package to say a whitelist in the first non-scl releaser.

ehelms:
Updated to try a different approach, this may be skirting or overlooking the actual issue but it appeared to fix the issue for me in all cases.

jlsherrill:
Can't comment on the correctness of the code, but it appears to solve the issue for me

dgoodwin:
Merging, thanks.

PR #190, opened 2015-07-24
Allow a user specific Copr remote SRPM URL.

awood:
By using a user specific URL, multiple members of the same team can
submit Copr releases without having to share credentials to the SRPM
remote location.  For example, each member of the team can set
COPR_REMOTE_LOCATION in .titorc to their <username>.fedorapeople.org
space.

The URL defined in the releaser will take precedence.


dgoodwin:
Looks good, thanks!

PR #195, opened 2015-09-04
Ask user to run DNF instead of YUM

FrostyX:
To get rid of:

    Yum-utils package has been deprecated, use dnf instead.
    See 'man yum2dnf' for more information.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -32,7 +32,7 @@ from tito.common import scl_to_rpm_option, get_latest_tagged_version, \
     get_commit_count, find_gemspec_file, create_builder, compare_version,\
     find_cheetah_template_file, render_cheetah, replace_spec_release, \
     find_spec_like_file, warn_out, get_commit_timestamp, chdir, mkdir_p, \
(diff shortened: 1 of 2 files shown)


jmrodri:
@dgoodwin Does tito still need to support OS versions without ```dnf```? If not, this patch looks sane. Otherwise, might need a little work to get it to know when to look for ```dnf``` vs ```yum```.

dgoodwin:
We do need to support OS versions without dnf, notably EL6 and EL7, and perhaps previous Fedora where it wasn't default, for the next few months anyhow. EL6-7 though aren't going anywhere soon.

FrostyX:
@dgoodwin Well, in this case we can examine whether the DNF is installed in the system and recommend the right package manager. It would take just about 5 lines of code. What do you think?

dgoodwin:
That would be fine yeah!

FrostyX:
Ok, I will write it on monday.

FrostyX:
Can you review it @dgoodwin, please?

dgoodwin:
Looks good, thank you!

PR #198, opened 2015-09-10
Add possibility to upload SRPM directly to Copr

FrostyX:
Additionally to the current `CoprReleaser` behavior in which the builded SRPM is uploaded to remote server and then submitted to the Copr via URL, I have implemented direct approach for submitting the package from your local storage right into the Copr. This feature is supported since `python-copr-1.58-1`.

With this patch it's supported the old style configuration:
    
    [copr]
    releaser = tito.release.CoprReleaser
    project_name = foo
    upload_command = scp %(srpm)s user@example.com:/home/user/public_html/
    remote_location = http://example.com/~user/

as well as the:

    [copr]
    releaser = tito.release.CoprReleaser
    project_name = foo

Can you review it @xsuchy and give me your blessing? :-)

### DIFF
--- a/releasers.conf.5.asciidoc
+++ b/releasers.conf.5.asciidoc
@@ -144,7 +144,7 @@ but if it is not present name of package directory is used.
 This releaser assume that you have ~/.oscrc correctly configured.
 
 tito.release.CoprReleaser::
-This releaser publish your src.rpm on internet and submit it to Copr.
+This releaser submits your src.rpm to Copr.
 
  [my-copr]
  releaser = tito.release.CoprReleaser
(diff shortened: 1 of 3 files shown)


dgoodwin:
Whoah really? This would be awesome.

xsuchy:
Looks good to me. Maybe - can you edit man page of releasers.conf(5) to state that these two options are optional now?

rholy:
I wonder why you rather use `copr-cli` than `python-copr` but this is probably not a good place for this discussion.
Anyway, I have no idea why you prefer `Popen` over `run_command` but I think that you should still check the exit status of the process.

dgoodwin:
As far as I'm concerned I would have no objections to python-copr usage.

xsuchy:
I thought about it as well, but it have several consequences. If we use python-copr, then we have to hard require python-copr in spec file (or put a lot of try/except in code). And since this is rather soft deps I do not think we should force everyone to install it.
And I hesitate to put Suggest into spec as we still support EL6, which does not recognize these tags. Even if you wrap it with %if/%endif.

domcleal:
I don't mind a hard dep on python-copr as it appears to just add a single package on a regular installation, not an entire tree of dependencies (like git-annex did).

FrostyX:
@rholy 
> Anyway, I have no idea why you prefer Popen over run_command but I think that you should still check the exit status of the process.

Well I had a problem with `run_command`. Even when I used `run_command_print` it wasn't showing the `copr-cli build ...` output on the fly. Which was kind of uncomfortable because you couldn't tell whether the tito froze or it is just still building. Did I do something wrong?

Next, will the process exit status help us with something? I was thinking about it, but errors are covered even now. See:

    Submiting build into Copr.

    Something went wrong:
    Error: Copr with name bar does not exist.

rholy:
Does tito exit with a non-zero exit status then? I guess no.

FrostyX:
@rholy, now it sets exit code by calling `error_out`, which has `1` on error.

@\everyone, do you think that any other enhancements should be done? Can I ask you to try the patch, whether it works for you as expected? Thank you.

dgoodwin:
Merged, thanks.

FrostyX:
Thank you for merging,
please let me know if some related issue appears.

PR #201, opened 2015-09-23
Fix changelog format function name

Allda:
Use _changelog_format() instead of _changelog_email() which no longer
exists.


dgoodwin:
Merged, thanks!

PR #203, opened 2015-11-11
Added ability to pass extra copr-cli build options to the copr releaser.

twiest:
This adds the ability to add 'copr-cli build' options to the releasers.conf file for the copr releaser.

My team will use this specifically for --nowait so that we can submit builds and move on.

This could be useful for the other options, like --memory, --chroot, --timeout, etc.

Here is an example releasers.conf file that works:
```
[copr-openshift-tools]
releaser = tito.release.CoprReleaser
project_name = openshift-tools
build_options = --nowait
```

Not specifying build_options in the config file also works (it's completely optional).

### DIFF
--- a/releasers.conf.5.asciidoc
+++ b/releasers.conf.5.asciidoc
@@ -151,12 +151,14 @@ This releaser submits your src.rpm to Copr.
  project_name = my-copr-project-name
  upload_command = scp %(srpm)s my.web.com:public_html/my_srpm/
  remote_location = http://my.web.com/~msuchy/my_srpm/
+ copr_options = --timeout 600
 
 Variables are:
 
  * project_name - this is name of your project on Copr
  * upload_command - this command is executed to upload src.rpm to internet. It can be scp, rsync or just cp. It may containt string "%(srpm)s" (even several times), which is substitued by real srpm path. (optional)
  * remote_location - how can be accessed the location above over www. (optional)
+ * copr_options - space separated options to pass directly to `copr-cli build`. Defaults to --nowait. (optional)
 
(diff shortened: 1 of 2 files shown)


dgoodwin:
Looks good, thank you for contributing.

Could I get a rename on the option, in tito we have builder's and releaser's, in this context Copr is a releaser (where we're releasing rpms). releasers.conf already has a mechanism for passing args to actual builders but in this case we're talking about a very diff object.

Because there's no required link between the releaser and a cli command, maybe we could just call it "copr_args" or "copr_options"? I'd be ok with "releaser_options" as well if you prefer.

Also if you were so inclined, I think it'd probably be ok to make --nowait the default behavior, it is for the very similar fedpkg releases and it's struck me as odd every time I've used it.

twiest:
@dgoodwin very reasonable suggestions, I'm definitely open to changing the name. I think copr_options works the best.

I also agree that I'd prefer to have --nowait as the default behavior.

I'll work on an update to the PR. Thanks for the feedback.

twiest:
@dgoodwin Please take a look and see if I implemented the changes like you were thinking.

I tested this with --dry-run, and the command looked good. Unfortunately copr seems to be down right now, so I wasn't able to test against the real thing.

twiest:
And, just as I said that, copr came back up. I just tested it and it works with copr too.

twiest:
@dgoodwin my team would like to start using this, but we don't want to use it until it's at least in master. 

Is there anything else I need to do to this PR to get it ready to be merged?

Thanks.

dgoodwin:
Sorry lost track of this, you are merged, thanks for the contrib.

twiest:
@dgoodwin great, thanks!

PR #209, opened 2016-01-19
Ensure GitAnnexBuilder cleanup returns to proper directory

ehelms:
In some cases, we aren't always assured that the releaser will return
us back to proper directory to re-lock the git annex file. This is
the case with the DistGitBuilder for example. This ensures we store
and return to the proper directory before issuing the lock on cleanup


dgoodwin:
Merging, thanks!

ehelms:
I found out minutes ago that for some releasers this will lead to errors where self.cur_cwd isn't always defined.

PR #211, opened 2016-01-25
Tagging with --use-version did not work with Mead projects.

awood:
An oversight on my part during all the Mead work.


dgoodwin:
Merged, thanks!

PR #217, opened 2016-03-25
Allow defining a custom changelog message

ehelms:


jmrodri:
You can supply a custom message by using the editor that appears after the changelog is generated. Is there a reason to have one passed in? Is it for automation purposes?

ehelms:
Correct - I want to automate tito tag with a custom changelog.

ehelms:
Any thoughts on this? I could use this change in a release so that my automation and team members can make use of it outside of my github fork.

share/tito_completion.sh
```diff
@@ -61,6 +61,7 @@ __tito_report_opts='
 __tito_tag_opts='
     --accept-auto-changelog
     --auto-changelog-message=
+    --changelog-message
```
dgoodwin: I'm tempted to go with just --changelog despite the consistency with the option above this.

src/tito/tagger/main.py
```diff
@@ -243,10 +239,21 @@ def _make_changelog(self):
 
                 write(fd, header)
 
-                for cmd_out in output.split("\n"):
-                    write(fd, "- ")
-                    write(fd, "\n  ".join(textwrap.wrap(cmd_out, 77)))
+                # don't die if this is a new package with no history
+                if self._changelog_msg is not None:
+                    write(fd, self._changelog_msg)
```
dgoodwin: Is the idea here that this is just a single line? If so, we should probably write it in with a leading "- ", it might be nicer for the user not to have to format it themselves.

If you're aiming for multi-line formatted changelogs, we should probably be looking at --changelog-file instead.

tito.8.asciidoc
```diff
@@ -101,6 +101,11 @@ Automatically accept the generated changelog.
 Use 'MESSAGE' as the default changelog message for new
 packages
 
+--changelog-message='MESSAGE'::
+Use 'MESSAGE' as the changelog when tagging, this can be specified
+everytime a package is tagged. Useful when needing to generate
+custom changelogs.
+
```
dgoodwin: Could you clarify this overrides the auto-generated changelog, and optionally if you'd prefer --changelog as well, update throughout. (I'll go either way)

dgoodwin:
Love the feature, definitely will merge, just a couple questions/comments, and if possible I would really be happy to see a unit test.  test/functiona/singleproject_tests.py would be well setup to plug in a test for this. I have a tendancy to break features I don't use without realizing it.

share/tito_completion.sh
ehelms: +1 works for me, will change to that

src/tito/tagger/main.py
ehelms: Could be single or multiline, I was going with the idea that if you are providing your own changelog message you don't want tito doing any assumed formatting. 

I could add '--changelog-file' as another option for those that have a file or want to write to a file first. In my current usage, I have a script that analyzes a git repo, pulls out commit messages, formats them the way I want and then calls `tito tag --changelog="MYMSG"` so enforcing a file for me would just cause me to have to write a file I don't need.

ehelms:
I will definitely add a unit test, it crossed my mind, I just wasn't sure where to put it. I will comment back here once I have pushed updates for things so far since there are some open conversation.

src/tito/tagger/main.py
dgoodwin: Would you be willing to go to --changelog "line1" --changelog "line2" etc? I'm not crazy about a --changelog "- line1\n- line2". It's fine programatically but very awkward for normal CLI user. I'd prefer this, or --changelog-file for programatic usage.

ehelms: Is there automatic support for multi-args or an example you know of off hand? I think adding '--changelog' as a multi-option will be a fine approach as for my usage I can still send a block with '--changelog'.

dgoodwin: There is yes, example here: https://github.com/dgoodwin/tito/blob/master/src/tito/cli.py#L448 Just comes out as a standard list.

Indeed you could just pack yours into one msg, just do some magic in the code to add a leading "- " if one isn't present on each --changelog when it's written if you don't mind. That way you can pre-format and pack into a line, and CLI users can just do --changelog "I did a thing." --changelog "I did another thing."

ehelms:
Updated with:

  1) support for multiple --changelog commands
  2) automatic addition of '- ' if not already present
  3) added tests to try to capture these different scenarios for use with the command

dgoodwin:
This is great, merging, thanks!

ehelms:
Thanks @dgoodwin ! if you need me to test the eventual RPM please let me know

Issue #218, opened 2016-04-12
tito doesn't support %autosetup/%autopatch


ignatenkobrain:
if I use `%autosetup` in spec file which totally correct building fails.

dgoodwin:
More details please. Sample spec file, output of the failure when run with --debug.

ignatenkobrain:
It's very easy.

```
%prep
%autosetup
```

It will work as (applying all patches):
```
%prep
%setup -q
%autopatch
```

`%autopatch` just applies all patches.

tito doesn't modify spec in case of autosetup, because it's simply not implemented.

ignatenkobrain:
so with `%autosetup` it tries to cd into `%{name}-%{version}` which doesn't exist in archive prepared by tito.

dgoodwin:
Full output of a build with --debug and --no-cleanup please, it would be great if you could go into the directory left over from the use of --no-cleanup and provide a full listing listing of the files/folders there.

Can you get it working with tito if you switch to traditional use of just %setup? What does your spec file look like in both cases.

http://www.rpm.org/wiki/PackagerDocs/Autosetup seems to indicate it should take the same arguments as %setup, and tito requires the -n option for setup as it does have to make assumptions about how to package the source.

ignatenkobrain:
> Can you get it working with tito if you switch to traditional use of just %setup? What does your spec file look like in both cases.

yes, it starts working

> http://www.rpm.org/wiki/PackagerDocs/Autosetup seems to indicate it should take the same arguments as %setup, and tito requires the -n option for setup as it does have to make assumptions about how to package the source.

if you will add hook somewhere in tito code to add `-n` to `%autosetup` it will start working

dgoodwin:
Is there a particular reason you've ignored 2 requests for a run with --debug, and a sample working vs not working spec file? 

Are you saying it works with just:

%setup

But fails with just:

%autosetup

But works again with:

%autosetup -n?

What options are you providing to -n?

If the fix is just to use -n, can you modify your spec file so it works? I use %setup -n in everything, including tito.spec.

ignatenkobrain:
```
[ignatenko@ignatenko dnf]$ tito build --test --rpm --debug
---- Builder class is None
Importing tito.builder.Builder
Using builder class: <class 'tito.builder.main.Builder'>
Building in temp dir: /tmp/tito/rpmbuild-dnf68vwt26w
[...]

PR #219, opened 2016-04-14
add support for %autosetup

ignatenkobrain:
Reference: https://github.com/dgoodwin/tito/issues/218
Signed-off-by: Igor Gnatenko <ignatenko@redhat.com>

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -755,6 +755,8 @@ class UpstreamBuilder(NoTgzBuilder):
                 patch_apply_index = array_index + 2
             elif line.startswith("%setup"):
                 patch_apply_index = array_index + 2  # already added a line
+            elif line.startswith("%autosetup"):
+                patch_apply_index = -1  # autosetup will do this for us
 
             array_index += 1
 
@@ -803,7 +805,8 @@ class UpstreamBuilder(NoTgzBuilder):
 
         lines.insert(patch_insert_index, "Patch%s: %s\n" % (patch_number,
             patch_filename))
-        lines.insert(patch_apply_index, "%%patch%s -p1\n" % (patch_number))
+        if patch_apply_index > 0:
+            lines.insert(patch_apply_index, "%%patch%s -p1\n" % (patch_number))
         self._write_spec(lines)
 
     def _write_spec(self, lines):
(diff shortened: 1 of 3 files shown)


ignatenkobrain:
@dgoodwin this is what I tried to say in issue. I didn't tested this patch, but if it will not work you will know what I meant

ignatenkobrain:
in simple case with `%autosetup` this patch works. I didn't tested it with `%patch` case, because I don't know how to reproduce it

dgoodwin:
Thanks it does not quite work, I had one in progress from yesterday quite similar that was also failing the same way, I will get them combined somehow and should have it working soon.

ignatenkobrain:
hm, can you show me what's the problem with my patch?

dgoodwin:
It looks like with %autosetup, if you build with --test and --no-cleanup, the resulting spec comes out with something like %autosetup -n tito-%{version} -n tito-correctversion, the build actually works but it's not correctly taking out the pre-existing -n. 

And with %setup as is the test spec file seemed to come out with a trailing -p1. I'm not quite sure why that happened yet.

ignatenkobrain:
@dgoodwin problem with appending `-p1` fixed by https://github.com/dgoodwin/tito/pull/219/commits/d802362964ae28f2610a3c952ba4a006bb26496f

ignatenkobrain:
@dgoodwin just tested tito from my branch:

After `tito build --test --srpm --no-cleanup` I see that bug, I guess it's only one remaining bug here

dgoodwin:
Yeah just that one remaining issue but at least the autosetup build will pass. 

%autosetup -n tito-%{version} -n tito-git-3.20362dd -p1

dgoodwin:
I should get back to my regular workload but I will take a look at that last issue hopefully before long unless you get it figured out first. It might be nice to extract the setup macro processing into another little function we could unit test in test/unit/ very quickly.

dgoodwin:
Your commits have been manually merged, I added some tests and a fix for the regex.

Thanks for the contribution!

ignatenkobrain:
@dgoodwin thanks! Can you apply this patches to F23+?

dgoodwin:
Builds submitted and in need of karma: https://bodhi.fedoraproject.org/users/dgoodwin

PR #224, opened 2016-05-13
better release number for untagged packages

xsuchy:
If package is unttaged and you build it with --test then previously 0 was used as number of commits
since last tag.
This results in packages which has no upgrade path.
With this patch it use number of commits since initial commit, which is growing number.
Therefore packages can be upgraded without usage of --force.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -879,6 +879,15 @@ def get_commit_count(tag, commit_id):
 
     if status != 0:
         debug("git describe of tag %s failed (%d)" % (tag, status))
+        debug("going to use number of commits from initial commit")
+        (status, output) = getstatusoutput(
+            "git rev-list --max-parents=0 HEAD")
+        if status == 0:
+            # output is now inital commit
+            (status, output) = getstatusoutput(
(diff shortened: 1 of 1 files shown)


domcleal:
:+1: nice improvement, this regularly catches me out.  Thanks for fixing it!

The only further problem I have with it is that the generated release is higher than the release if the package is then tagged with `--keep-version` (since Release doesn't really need incrementing on the first build).  With version at 0.4.0, release at 1, and `tito build --srpm --test` on this untagged package will generate an EVR of: `0.4.0-1.git.2768.14a280d.fc22`, which is newer than `0.4.0-1.fc22`.  My only thought to fix that would be to also decrement the existing release number by one.

xsuchy:
Aha. I either never tag it (quick'n'dirty experiments) or I properly tag it with version increment. 

In your example if 0.4.0-1.fc22 if already exist in distribution, then test builds will have lower version and will not upgrade. It is conflict of two use cases.

domcleal:
True!

Ablu:
Thank you. I just tested it and it works :)

PR #229, opened 2016-05-26
Allow customizing git commit message

### LINKED ISSUE #228: Allow changing git commit message
When creating a tag, tito will create a git commit. It would be nice to be able to change the commit message. Currently it is [hardcoded to use the "Automatic commit of package [X] Y [Z]." format](https://github.com/dgoodwin/tito/blob/77c02ceb441386850237756b45b3909574a0ae9e/src/tito/tagger/main.py#L481).

I would like to have an option to put custom format string into tito.props.

lubomir:
Fixes #228.

I did not find a better way to change the configuration in tests. Please let me know if I should rework it.

### DIFF
--- a/src/tito/tagger/main.py
+++ b/src/tito/tagger/main.py
@@ -19,8 +19,14 @@ import re
 import rpm
 import shutil
 import subprocess
+import sys
 import tempfile
 import textwrap
+
+try:
+    from shlex import quote
+except ImportError:
+    from pipes import quote
 
 from string import Template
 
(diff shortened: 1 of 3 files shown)


src/tito/tagger/main.py
```diff
@@ -478,9 +478,16 @@ def _update_package_metadata(self, new_version):
         run_command("git add %s" % os.path.join(self.full_project_dir,
             self.spec_file_name))
 
-        run_command('git commit -m "Automatic commit of package ' +
-                '[%s] %s [%s]."' % (self.project_name, self.release_type(),
-                    new_version_w_suffix))
+        fmt = ('Automatic commit of package '
+               '[%(name)s] %(release_type)s [%(version)s].')
+        if self.config.has_option(BUILDCONFIG_SECTION, "commit_message_format"):
```
dgoodwin: Could I just get a rename on this property to "tag_commit_msg_format", I know that's probably the only git commit we do but just incase for the future. Everything else looks good to me.

src/tito/tagger/main.py
```diff
@@ -478,9 +478,16 @@ def _update_package_metadata(self, new_version):
...
+            fmt = self.config.get(BUILDCONFIG_SECTION, "commit_message_format")
```
xsuchy: If you are reading this config, then you should escape that value. What if the message in config contains quotes?

dgoodwin: Good thinking, we should chuck a unit test in for that too.

lubomir:
Current version of patch has these changes:

 * Option rename to `tag_commit_message_format`. There is at least one other commit message when running `tag init`, but that one can easily be modified manually with `git commit --amend`, so I don't feel it needs to be customizable.
 * A test for format with a quote is added and code is fixed to work with such message.
 * There is better error message when unknown placeholder is used in the format. Test for this is there as well.

xsuchy:
:+1:

dgoodwin:
Merging, thanks lubomir!

PR #236, opened 2016-10-26
Don't use a special tagger for the `--use-version` case

stevekuznetsov:
Instead of special-casing the case where a user has supplied
the specific version to use, we can instead use the same version
tagger for both cases and detect inside of that tagger whether
or not the user has supplied the tag.

Signed-off-by: Steve Kuznetsov skuznets@redhat.com

### DIFF
--- a/src/tito/tagger/main.py
+++ b/src/tito/tagger/main.py
@@ -371,7 +371,7 @@ class VersionTagger(ConfigObject):
             mvn_new_version))
         run_command("git add %s" % pom_file)
 
-    def _bump_version(self, release=False, zstream=False, force=False):
+    def _bump_version(self, release=False, zstream=False):
         """
         Bump up the package version in the spec file.
 
@@ -405,7 +405,7 @@ class VersionTagger(ConfigObject):
                                         increase_zstream(match.group(2)),
                                         "\n"
                         ))
-                elif force:
+                elif hasattr(self, '_use_version'):
                     match = re.match(version_regex, line)
                     if match:
                         line = "".join((match.group(1),
@@ -656,19 +656,9 @@ class ReleaseTagger(VersionTagger):
 
 class ForceVersionTagger(VersionTagger):
     """
-    Tagger which forcibly updates the spec file to a version provided on the
-    command line by the --use-version option.
-    TODO: could this be merged into main taggers?
+    Legacy Tagger which is chosen when a user passes the `--use-version`
+    flag in on the command line. This implementation has been merged with
+    the default `VersionTagger`, and remains only for backward compatibility
+    for users that were overriding this class with a custom implementation.
+    TODO: remove this class in a future release
     """
-
-    def _tag_release(self):
-        """
-        Tag a new release of the package.
-        """
-        self._make_changelog()
-        new_version = self._bump_version(force=True)
-        self._check_tag_does_not_exist(self._get_new_tag(new_version))
-        self._update_changelog(new_version)
-        self._update_setup_py(new_version)
-        self._update_pom_xml(new_version)
-        self._update_package_metadata(new_version)


src/tito/tagger/main.py
```diff
@@ -652,23 +652,3 @@ def _tag_release(self):
     def release_type(self):
         """ return short string "minor release" """
         return "minor release"
-
-
-class ForceVersionTagger(VersionTagger):
```
dgoodwin: Causing a test failure due to an import in tagger/__init__.py.

src/tito/cli.py
```diff
@@ -654,9 +654,7 @@ def main(self, argv):
             error_out("Tagging has been disabled in this git branch.")
 
         tagger_class = None
-        if self.options.use_version:
-            tagger_class = get_class_by_name("tito.tagger.ForceVersionTagger")
-        elif self.config.has_option("buildconfig", "tagger"):
+        if self.config.has_option("buildconfig", "tagger"):
```
dgoodwin: So this definitely looks more logical, my only concern is changing behavior that might catch someone off guard. Right now anyone using one of the other three bundled taggers, or a custom one in their repo, *might* be using --use-version and expecting it to override their normal versioning mechanism but instead the normal tagger runs and does nothing with that info.

So any ideas how we could roll it out as smoothly as possible? 

We could add a "supports_manual_version" method on the tagger base class, default it to False there, then set to true just for the main VersionTagger, and then make absolutely sure an attempt to use another tagger with --use-version will fail, and we could add support to other taggers as needed, and anyone who tries will get a clear error about it.

A unit test to this effect would be awesome too.

I'm open to other options though but I would like to find someway to make sure we don't trigger the above scenario.

src/tito/tagger/main.py
```diff
@@ -405,7 +405,7 @@ def _bump_version(self, release=False, zstream=False, force=False):
                                         increase_zstream(match.group(2)),
                                         "\n"
                         ))
-                elif force:
+                elif self._use_version is not None:
```
stevekuznetsov: @dgoodwin I am relatively certain that this logic is identical to the previous logic -- 

The only diff between the `VersionTagger` implementation of `_tag_release` and the `ForceVersionTagger` implementation is that the `ForceVersionTagger` is passing `force` in as `True`. If there exists a `self._use_version`, this is identically the case when the `ForceVersionTagger` would have been in use anyway, so as far as I can tell, if we change the check on the `VersionTagger` to simply look for the existence of the `_use_version` field we have the same behavior.

stevekuznetsov: If people were able to over-ride the `ForceVersionTagger` in downstream implementations of custom taggers, however, we need to keep that class around (with no methods?) so we don't break them.

dgoodwin: Am I correct that a user who has a custom tagger defined today (a custom tagger for example), if they are using --use-version their tagger implementation gets changed on the fly and uses ForceVersionTagger, resulting in the specific version they set being tagged.

With this change, it looks like if they kept doing this it will now just call their normal configured tagger by default, which probably ignores the version specified but silently works and does it's normal thing.

stevekuznetsov: > it looks like if they kept doing this it will now just call their normal configured tagger by default,

Yes

> which probably ignores the version specified

No, the line of code that I am commenting on in this thread is what makes the original `VersionTagger` keep the force behavior.


I will rework this PR to have a no-op `ForceVersionTagger` so we don't break anyone who is implementing their own.

stevekuznetsov:
@dgoodwin I have updated the PR

src/tito/tagger/main.py
dgoodwin: I am ok with removing ForceVersionTagger, that breakage will be clearly visible to anyone that hits it so please feel free to drop that and adjust the places it's imported.

The one issue I'm hoping to avoid is a customer tagger out in the wild, which does not subclass VersionTagger. Suddenly they do a tag --use-version and it no longer does what it did before (because the override to ForceVersionTagger is gone), and just silently runs their normal tagger as if they didn't --use-version. 

I like the goal you've got here to start letting the taggers themselves decide what to do with --use-version I just want to make sure we don't silently alter any behavior on anyone. This is why I'm thinking we should implement a "def supports_use_version(self)" or equivalent, and have it set to true on VersionTagger. We would check if --use-version is set in the CLI code, check that method on that tagger, and throw an error if it's not supported. Then in the scenario above we get a clear error that you can't use --use-version with tagger X and why, and they know they need to adjust their tagger now.

dgoodwin: My apologies I missed how the diff changed, so now we *do* still swap in the ForceVersionTagger if you specify --use-version, it's just a passthrough straight to VersionTagger? This looks to avoid what I was worried about, thanks!

dgoodwin: Looks like test failures here in undo tag tests:

```
RROR: test_tag_with_custom_message (functional.singleproject_tests.SingleProjectTests)                                                                                     [232/4582]
----------------------------------------------------------------------
  File "/home/dgoodwin/src/tito/test/functional/singleproject_tests.py", line 109, in test_tag_with_custom_message
    tito("tag --accept-auto-changelog")
  File "/home/dgoodwin/src/tito/test/functional/fixture.py", line 103, in tito
    return CLI().main(argstring.split(' '))
  File "/home/dgoodwin/src/tito/src/tito/cli.py", line 202, in main
    return module.main(argv)
  File "/home/dgoodwin/src/tito/src/tito/cli.py", line 673, in main
    return tagger.run(self.options)
  File "/home/dgoodwin/src/tito/src/tito/tagger/main.py", line 111, in run
    self._tag_release()
  File "/home/dgoodwin/src/tito/src/tito/tagger/main.py", line 132, in _tag_release
    new_version = self._bump_version()
  File "/home/dgoodwin/src/tito/src/tito/tagger/main.py", line 408, in _bump_version
    elif self._use_version is not None:
AttributeError: 'VersionTagger' object has no attribute '_use_version'
```

stevekuznetsov:
@dgoodwin OK, sorry -- was waiting on the design review to figure out how to run the tests. Would be great if the Docker sandboxes worked. I've updated the `is not None` check to be `hasattr` check instead. I am still getting this error but it looks like it's not from my changes:
```
======================================================================
FAIL: test_conformance (unit.pep8_tests.TestPep8)
----------------------------------------------------------------------
  File "/home/skuznets/Documents/go/src/github.com/dgoodwin/tito/test/unit/pep8_tests.py", line 70, in test_conformance
    self.fail("Found PEP8 errors that may break your code in Python 3:\n%s" % output)
AssertionError: Found PEP8 errors that may break your code in Python 3:
1       E731 do not assign a lambda expression, use a def
-------------------- >> begin captured stdout << ---------------------


Testing in: /home/skuznets/Documents/go/src/github.com/dgoodwin/tito/test/unit/../..

/home/skuznets/Documents/go/src/github.com/dgoodwin/tito/test/unit/../../src/tito/common.py:218:9: E731 do not assign a lambda expression, use a def

--------------------- >> end captured stdout << ----------------------

----------------------------------------------------------------------
```

dgoodwin:
Yeah that one is fine that's just spontaneously appeared via changes to the rpms that do the pep8 checking. 

Thanks for fixing this, merging now.

PR #238, opened 2016-11-22
Cargo

msehnout:
**State** of this PR: WIP

**Usage** in tito.props:
```ini
[remote_pkg_manager]
type = cargo
config_file = Cargo.toml
```
How it works in the code:
* New step in _tag_release method. Here I read the config file and decide what tagger to use. (_update_pkg_managers method)
* New file with the tagger class for Cargo (adding other pkg managers would require similar class).
* Tests for this class.

**Considerations**: Do we want to include a whole [toml](https://github.com/toml-lang/toml) parser?


msehnout:
Do you have any comments so far?

dgoodwin:
Sorry haven't had time to look yet but I should be able to this week.

jmrodri:
Do we really need a special tagger for Cargo? Why not use the template mechanism? We did the following to update a version.rb file which was used by the gemspec to get the version. 

```
[version_template]
template_file = ./server/version.rb.tmpl
destination_file = ./lib/fusor/version.rb
```

I would guess a cargo.toml.tmpl that could then be updated and copied over. Or does the cargo.toml require other changes besides the version?

src/tito/tagger/main.py
```diff
@@ -132,10 +133,42 @@ def _tag_release(self):
         new_version = self._bump_version()
         self._check_tag_does_not_exist(self._get_new_tag(new_version))
         self._update_changelog(new_version)
+        # Update pkg managers used by the project (e.g. Cargo, Cabal)
+        # XXX: merge setup_py and pom_xml into this method ?
+        self._update_pkg_managers(new_version)
         self._update_setup_py(new_version)
         self._update_pom_xml(new_version)
         self._update_package_metadata(new_version)
```
jmrodri: Not specific to this PR, but I'm thinking it might be time to rethink this portion of the VersionTagger.  Right now we're updating specific files: setup.py, pom.xml, etc. Would be better if we had some sort of registry so this code would change to :+1: 

```
self._update_SOME_GOOD_NAME(new_version)
```

That method could look through a registry which was configured in the tito.props.

```
def _update_SOME_GOOD_NAME(version):
   for file in self.registry:
      DO THE RIGHT THING :)
```

msehnout:
@jmrodri We've already been discussing the troubles with the template mechanism. The main problem is, that you might want to change the content of the `Cargo.toml` file without introducing a new tag. In that case you need to change the content of both `Cargo.toml` and `Cargo.tmpl` (which is basically the same except the part with version) so you can build your project using Cargo and your changes won't be overwritten next time you make a tag. Moreover I don't think that this patch adds too much complexity, so its worth the convenience gain.

src/tito/tagger/cargotagger.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class CargoTagger:
+    """
+    Cargo is package manager for the Rust programming
+    language: http://doc.crates.io/manifest.html
+    It uses Cargo.toml file as its configuration file.
+    XXX: I'm not including a Toml parser, because I
+    don't want to introduce new dependencies.
+    """
```
dgoodwin: This class should be renamed and *maybe* relocated, it's not actually a tagger in that it doesn't adhere to the "interface" a tagger would be expected to adhere to. IMO this is more of a utility method.

test/unit/common_tests.py
```diff
@@ -261,6 +262,57 @@ def test_renders_cheetah_missing_result(self, mock_tempfile, mock_run_command, m
...
+
+        CargoTagger.tag_new_version("3.3.3-1", self.config_file)
+        output = open(self.config_file, 'r').readlines()
+
+        self.assertEquals(11, len(output))
+        self.assertEquals("version = \"3.3.3\"\n", output[3])
+        self.assertEquals("regex = \"1.0.0\"\n", output[7])
+        self.assertEquals("version = \"0.1.0\"\n", output[10])
+
+
```
dgoodwin: Could I get one more test going through the entire spec file + version tagger workflow, with a rust Config.toml in the working dir as well? Just want to make absolutely sure we can keep this working for you amidst future changes.

src/tito/tagger/cargotagger.py
```diff
@@ -0,0 +1,60 @@
+import re
+import os
+
+from tito.common import debug
+
+
+class CargoTagger:
```
dgoodwin: This class should be renamed and maybe relocated, it's not actually a tagger in that it doesn't adhere to the "interface" a tagger would be expected to adhere to. IMO this is more of a utility method.

src/tito/tagger/cargotagger.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Cargo is package manager for the Rust programming
+    language: http://doc.crates.io/manifest.html
+    It uses Cargo.toml file as its configuration file.
+    XXX: I'm not including a Toml parser, because I
+    don't want to introduce new dependencies.
```
dgoodwin: Would python's INI parser read this sensibly? Looking at your tests it almost looks like it would.

src/tito/tagger/main.py
dgoodwin: I forgot about POM, yeah we might be approaching a point where we should design something a little better for these language specific version bumpers.

src/tito/tagger/main.py
```diff
@@ -132,10 +133,42 @@ def _tag_release(self):
...
+                    file_name = self.config.get(section_name, option_file)
+                # Run update based on the type of the project
+                debug('Found these options: name=' + file_name + ', type: ' + project_type)
+                if project_type == "cargo":
+                    if file_name is None:
+                        file_name = "Cargo.toml"
+                    CargoTagger.tag_new_version(new_version, os.path.join(self.full_project_dir, file_name))
+                else:
+                    # TODO: write a list of supported managers somewhere (man page, --help)
+                    debug('Unknown project type: ' + project_type)
```
dgoodwin: This feels like an error here, if this is misconfigred it's time to exit IMO.

src/tito/tagger/main.py
```diff
@@ -132,10 +133,42 @@ def _tag_release(self):
...
+                    return
+                # Add config file into the index
+                run_command("git add %s" % file_name)
+            else:
+                debug('Section ' + section_name + ' was specified, but no ' + option_type + ' was found.')
```
dgoodwin: Perhaps another error condition?

src/tito/tagger/main.py
```diff
@@ -132,10 +133,42 @@ def _tag_release(self):
...
+        # XXX: merge setup_py and pom_xml into this method ?
+        self._update_pkg_managers(new_version)
         self._update_setup_py(new_version)
         self._update_pom_xml(new_version)
         self._update_package_metadata(new_version)
 
+    def _update_pkg_managers(self, new_version):
+        section_name = 'remote_pkg_manager'
+        option_type = 'type'
+        option_file = 'config_file'
```
dgoodwin: Question for you, why drive this by config vs just detecting a Config.toml, and assuming that means we should try to update a version in it? (I believe this is what we do for setup.py) Is the config helpful or would it be safe to just automatically do it. (perhaps failing gracefully if there's no version in there for some reason)

If config is better, remote_pkg_manager feels ill-named, what is the remote referring to? Things like "pkg_metadata" are coming to mind if we were to apply this config to setup.py and pom.xml. (though we can't really as that would change automatic behaviour for users now, might be another indication we should think about making this happen automatically)

src/tito/tagger/cargotagger.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    @staticmethod
+    def tag_new_version(new_version, config_file):
+        """
+        Find the line with version number  and change
+        it to contain the new version.
+        """
+        if not os.path.exists(config_file):
+            debug('Config file was not found.')
```
dgoodwin: Feels like an error as well.

msehnout:
Thanks for your comments!

Just to sum up:
 * 1 more test, that checks both the spec file and config file
 * Don't introduce any section in tito.props but do it automatically instead (as it does for Python and Java)
 * Throw more errors :-) 

Suggestion:
What if I wrote something like this:
```python
         self._check_tag_does_not_exist(self._get_new_tag(new_version))
         self._update_changelog(new_version)
         CargoBump.tag_new_version(project_path, new_version)
         self._update_setup_py(new_version)
         self._update_pom_xml(new_version)
```
and you can then rewrite the section as you wish, e.g. the registry. Would this be ok with you?

dgoodwin:
Yeah if you're in agreement, I think I'd be ok with it happening automatically, though if so we may not want all those extra errors I was calling for. :) In that case the user didn't explicitly set up config which is failing, so feel free to skip any of those calls for errors that no longer make sense if we're going this route.

The example you gave above is fine with me, registry refactoring is definitely not something required for this PR.

jmrodri:
@dgoodwin @msehnout agreed registry refactoring not required for this PR. I just wanted to bring it up since we seem to keep adding language specific updates which starts to smell :)

@msehnout the example with CargoBump on the surface looks ok.

msehnout:
The `CargoBump` class should be complete, but I'm not sure where to put the complete test and also how to write it. Is it supposed to be something like the `SingleProjectTests`?

Basically I would like to write something like this, but in Python:
```bash
#!/bin/bash

TEST=$RANDOM
# Create new cargo package
mkdir $TEST
cd $TEST
cat << EOF > Cargo.toml
[package]
name = "testing_pkg"
version = "0.0.1"
authors = ["Nobody <test@tito.com>"]

[dependencies]
EOF
cat << EOF > $TEST.spec
Name:           hello_tito
Version:        0.1.7
Release:        1%{?dist}
Summary:        testing package

License:        Public Domain
Source0:        hello_tito-%{version}.tar.gz

%description


%prep
%autosetup


%build

%install

%files

%changelog
EOF

sed -i "s/hello_tito/$TEST/" $TEST.spec
git init
git add .
git commit -m "test"
tito init
tito tag --accept-auto-changelog
git show HEAD
```

dgoodwin:
Yeah single project tests is the example I would work off and a good place for the test to live as well, you can do a separate test module if you prefer. In there you should see examples for creating a stub project, you'd just need to adapt for your cargo file.

msehnout:
I've finally written the test class. It creates a directory with 'Cargo.toml' inside and checks that the file contains the desired version and it's part of the last commit.

dgoodwin:
Getting a couple test failures, any ideas?

```
======================================================================
FAIL: test_complicated_case (unit.common_tests.CargoTransformTest)
----------------------------------------------------------------------
  File "/home/dgoodwin/src/tito/test/unit/common_tests.py", line 311, in test_complicated_case
    self.assertEquals("version = \"3.3.3\"\n", output[3])
AssertionError: 'version = "3.3.3"\n' != 'version = "2.2.2"\n'
-------------------- >> begin captured stdout << ---------------------
Cargo.toml file not found, this is probably not a Rust project

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: test_simple_case (unit.common_tests.CargoTransformTest)
----------------------------------------------------------------------
  File "/home/dgoodwin/src/tito/test/unit/common_tests.py", line 288, in test_simple_case
    self.assertEquals("version = \"2.2.2\"    # the current version, obeying semver\n", output[3])
AssertionError: 'version = "2.2.2"    # the current version, obeying semver\n' != 'version = "0.1.0"    # the current version, obeying semver\n'
-------------------- >> begin captured stdout << ---------------------
Cargo.toml file not found, this is probably not a Rust project

--------------------- >> end captured stdout << ----------------------
```

msehnout:
Ouch, I changed the API of CargoBump class. I will fix them.

msehnout:
It should work now.

dgoodwin:
Yup we good, thanks @msehnout

PR #242, opened 2016-12-13
explain how automatic tagging was done

xsuchy:
This will create description as:

    Automatic commit of package [tito] release [0.6.9-1].
    
    Created by command:
    
    /usr/bin/tito tag

It use the feature of git-commit that additional -m option are used as next paragraph.

This can help newbies to identify tool and process how tagging was actually done (for example usage of --keep-version).


dgoodwin:
Cool, thanks.

PR #245, opened 2017-01-05
Relabel test container volume contents for SELinux

stevekuznetsov:
To allow the tests to run from a host with SELinux enabled, the data
needs to be relabeled to the correct SELinux contexts. Adding the `Z`
label to the volume mount allows for this to occur.


stevekuznetsov:
@dgoodwin PTAL

PR #248, opened 2017-01-05
Use `dnf` instead of `yum` in Fedora Rawhide tests

stevekuznetsov:


stevekuznetsov:
@dgoodwin PTAL

PR #259, opened 2017-01-24
Collapse tagger class selection logic

stevekuznetsov:
In 26b02a7, the utility of the `ForceVersionTagger` was merged with the
base `VersionTagger`, so it is no longer necessary to ever explicitly
invoke the `ForceVersionTagger` class. Furthermore, the `elif` and
`else` branches of the decision tree for the tagger class were identical
in the previous implementation, so they were collapsed as well. The
remaining tagger class assignment is a simple query of the `tito.props`
file.


stevekuznetsov:
@dgoodwin PTAL

PR #260, opened 2017-01-24
Added an option to not escalate privileges on `tito build --install`

stevekuznetsov:
When using `tito build --install` on a system without `sudo`, such as
inside of a Linux container build, the current implementation fails.
Furthermore, if the user is running `tito build --install` with the
correct level of privilege, it is not necessary to escalate privileges
further for the installation step.

This patch adds the `--dont-escalate-privileges` flag to `tito build`
and defaults it to `True`, which keeps the behavior backwards compat-
ible. Users will want to use this flag when building RPMs inside of a
container or when running `tito build` with the requisite permissions
for installing in the first place.

### DIFF
--- a/src/tito/cli.py
+++ b/src/tito/cli.py
@@ -307,6 +307,10 @@ class BuildModule(BaseCliModule):
                 action="store_true", default=False,
                 help="Install any binary rpms being built. (WARNING: " +
                     "uses sudo rpm -Uvh --force)")
+        self.parser.add_option("--no-sudo", dest="escalate",
+                action="store_false", default=True,
+                help="Don't escalate privileges when installing. Use when " +
+                "running this command with required privileges.")
         self.parser.add_option("--dist", dest="dist", metavar="DISTTAG",
                 help="Dist tag to apply to srpm and/or rpm. (i.e. .el5)")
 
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
@dgoodwin how does this look?

src/tito/cli.py
```diff
@@ -307,6 +307,10 @@ def __init__(self):
                 action="store_true", default=False,
                 help="Install any binary rpms being built. (WARNING: " +
                     "uses sudo rpm -Uvh --force)")
+        self.parser.add_option("--dont-escalate-privileges", dest="escalate",
+                action="store_false", default=True,
+                help="Don't escalate privileges when installing. Use when " +
+                "running this command with required privileges.")
```
dgoodwin: Could we do a more concise --no-sudo or --nosudo, privilege escalation is a bit verbose and feels a bit like exploit terminology to me.

Otherwise it looks good just need to document the new option in the manpage.

stevekuznetsov: Will do

stevekuznetsov:
@dgoodwin tag renamed to `--no-sudo`, added entry to the manpage, also added one for `tito tag --use-release` which I guess we forgot to do last time.

dgoodwin:
Perfect ty!

PR #261, opened 2017-01-24
Factor out the version->tag mapping in the Builder

stevekuznetsov:
When a custom implementation of a tito Tagger is used, the mapping
function that is used to generate a tag from a version in the custom
Tagger must also be used in whichever tito Builder is used on that
repository. This commit factors out the mapping of version to tag so
that custom implementations of a Builder can inject custom logic here
with minimal effort, as is possible on the tito.VersionTagger by
overriding the _get_new_tag() function.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -437,13 +437,21 @@ class Builder(ConfigObject, BuilderBase):
                     "tag, building untagged test project")
                 build_version = get_spec_version_and_release(self.start_dir,
                     find_spec_like_file(self.start_dir))
-            self.build_tag = "%s-%s" % (self.project_name, build_version)
+            self.build_tag = self._get_tag_for_version(build_version)
 
         self.spec_version = build_version.split('-')[-2]
         self.spec_release = build_version.split('-')[-1]
         if not self.test:
             check_tag_exists(self.build_tag, offline=self.offline)
         return build_version
+
+    def _get_tag_for_version(self, version):
+        """
+        Determine what the tag will look like for a given version.
+        Can be overridden when custom taggers override counterpart,
+        tito.VersionTagger._get_new_tag().
+        """
+        return "%s-%s".format(self.project_name, version)
 
     def tgz(self):
         """


stevekuznetsov:
I'm not very happy with this patch -- this seems like a place where things are not perfectly factored. In a perfect world, the Builder in use would ask the Tagger in use for this information, right? This would allow for one canonical source of information, one place where the mapping from a version to a tag could be made. In the past, custom implementations were over-riding entire functions inside of the `tito.Builder` in order to change the one line in `_get_build_version()` that I changed. This patch would allow for custom implementations to not need to touch `_get_build_version()` at all, as they do not care about the rest of the function, just this mapping. In the future, though, maybe a more through path to restructure how the Builder determines the tag for a version would be useful.

@dgoodwin what are your thoughts?

dgoodwin:
This implies that if you're in this situation with a custom tagger, you now need a custom builder just to cover the customizations in your tagger?

That isn't great but I really like your more perfect world description, that would really open the door to a lot more use cases if the tag format wasn't so rigid and instead focused entirely on the tagger implementation. 

I see that builder doesn't really know anything about the tagger right now, however it could pretty easily get access to the class name and instantiate one I suspect. 

We init the tagger in cli.py, it looks like we could do that within the builder if we just assume keep_version and offline values, neither of which really matter in this context. Other possibility would be passing a tagger in to the builder when created, not sure how hard that would be.

stevekuznetsov:
> This implies that if you're in this situation with a custom tagger, you now need a custom builder just to cover the customizations in your tagger?

Yes, the custom tagger overrides `tito.VersionTagger._get_new_tag()`, so the mapping is not `name-version-suffix` but simply `version`. The tagger serializes the `version-suffix` part of the tag as the version in `.tito/packages` and the builder uses that to create the `name-version` tag it's looking for. Unless more information than just the version can be stored on disk, the builder will need to know how to correctly determine what the tag is for a version.

> that would really open the door to a lot more use cases if the tag format wasn't so rigid and instead focused entirely on the tagger implementation.

At the same time, though, I really don't know if:
 - the project ever meant to be less rigid
 - users are expected to always pair a custom tagger with a custom builder

If we keep the version->tag mapping encapsulated in the tagger, a user could potentially use a custom tagger only, no custom builder. 

In any case, that seems like a large architectural change and I'm not sure about it all. The current patch that is proposed simply allows users that have custom tagging implementations to have a single place in the builder where they can override the functionality and sync their tagger with their builder. Today they need to copy large functions like `_get_build_version()` while they're changing only one part, which leads to the rest of the function slowly drifting away from the real implementation in upstream.

dgoodwin:
Ok we can take this as is and try to improve later hopefully. I like where you were headed, tag format has been a sticking point for some.

PR #262, opened 2017-01-25
Flush output buffers

stevekuznetsov:
When using `$DEBUG=1`, often the output from different parts of `tito`
that output to different streams (for instance, `stderr` and `stdout`)
appears in the terminal in a garbled order. Furthermore, the stacktraces
generated by the Python interpreter are out of order with the output
from the `tito` process and often fragmented. By flushing the output
buffers on `tito.common._out()` we should be able to work around this.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -226,6 +226,8 @@ def _out(msgs, prefix, color_func, stream=sys.stdout):
     else:
         print(color_func(fmt % {'prefix': prefix, 'msg': msgs}), file=stream)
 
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
@dgoodwin this may have performance implications and is mostly useful for debugging, so I can gate the flush on the same `if 'DEBUG' in os.environ` stanza if you think it's necessary. Or if this is a frivolous patch, I can close. Just ran into a lot of garbled output.

PR #264, opened 2017-01-25
Added `--quiet` and `--verbose` to `tito build`

stevekuznetsov:
Refactor `rpmbuild` invocation for readability

---

Added `--quiet` and `--verbose` to `tito build`

The current output levels from `tito build` are both very verbose in the
case that nothing goes wrong and often missing full context when some-
thing does go wrong. By exposing `--quiet` and `--verbose`, the build
process should be able to modulate the amount of output coming from the
internal process and from `rpmbuild`.

---

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -77,6 +77,9 @@ class BuilderBase(object):
         self.scl = self._get_optional_arg(args, 'scl', [None])[0] or \
                 self._get_optional_arg(kwargs, 'scl', '')
 
+        self.quiet = self._get_optional_arg(kwargs, 'quiet', False)
+        self.verbose = self._get_optional_arg(kwargs, 'verbose', False)
+
         rpmbuildopts = self._get_optional_arg(args, 'rpmbuild_options', None)
         if rpmbuildopts:
             self.rpmbuild_options = ' '.join(rpmbuildopts)
@@ -232,6 +235,14 @@ class BuilderBase(object):
             return "--noclean"
         return "--clean"
 
+    def _get_verbosity_option(self):
+        if self.quiet:
+            return "--quiet"
+        elif self.verbose:
+            return "--verbose"
+        else:
+            return ""
+
     def rpm(self):
         """ Build an RPM. """
         self._create_build_dirs()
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
@dgoodwin it may be more valid to just pass these through the `--rpmbuild-options` but Clayton and I were thinking we could use them to further change which bits of the `tito` process would output on a `tito build` -- so `--quiet` would not _only_ be used to quiet the `rpmbuild` exec.

src/tito/builder/main.py
```diff
@@ -454,7 +468,7 @@ def _get_tag_for_version(self, version):
         Can be overridden when custom taggers override counterpart,
         tito.VersionTagger._get_new_tag().
         """
-        return "%s-%s".format(self.project_name, version)
+        return "{}-{}".format(self.project_name, version)
```
stevekuznetsov: @dgoodwin sorry, I had not run tests on #261 before you merged it and this was an incorrect mix of print-formatting styles.

src/tito/cli.py
```diff
@@ -332,6 +332,11 @@ def __init__(self):
                 help="Custom arguments specific to a particular builder."
                     " (key=value)")
 
+        self.parser.add_option("--quiet", dest="quiet", action="store_true",
+                help="Suppress output from the build process.")
+        self.parser.add_option("--verbose", dest="verbose", action="store_true",
+                help="Expose more output from the build process.")
```
dgoodwin: Need to add these to the manpages.

--verbose is a bit close to --debug, we should probably merge them at some point in the future but if we do I'd lean towards the new --verbose in any case.

src/tito/builder/main.py
dgoodwin: UNACCEPTABLE!

src/tito/cli.py
stevekuznetsov: Yeah, wasn't really sure what would be best -- since `rpmbuild` exposed both I figured we could too, to have three levels and keep the default the same as today.

stevekuznetsov:
@dgoodwin updated the `man` page.

dgoodwin:
Just needs a rebase and I'll merge.

stevekuznetsov:
@dgoodwin rebased

PR #265, opened 2017-01-26
Refactored version->tag mapping logic in Tagger

stevekuznetsov:
Only flush output stream if flushing is supported

While file handles support the `.flush()` method, some of the more
UNIX-ey pipes and tees do not. We should try to `.flush()` when we
can, but not bother when we can't.

---

Added support for choosing platforms for tests

Running the entire suite of containerized tests across all of the
platforms that are supported takes quite a while, especially if the
container images need to be built. Now, the platforms that are to be
tested can be specified with `$PY2_DISTROS` and `$PY3_DISTROS`.

---

Refactored version->tag mapping logic in Tagger

While the `tito.VersionTagger._get_new_tag()` method encapsulated some
of the version->tag mapping logic, other areas in the `VersionTagger`
used their own logic to do the mapping. This commit ensures that this
mapping happens in one place, which allows for custom implementations
to override the behavior simply.

---

### DIFF
--- a/hacking/runtests.sh
+++ b/hacking/runtests.sh
@@ -29,6 +29,10 @@
 # to run on the first build.
 # (or if you remove the images)
 #
+# To run tests on only one platform, set the environment
+# variables $PY2_DISTROS and $PY3_DISTROS, like so:
+# PY3_DISTROS= PY2_DISTROS=fedora-25 hacking/runtests.sh
+#
 #
 # Expected output resembles:
 #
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
@dgoodwin first commit fixed `tito tag --undo` for me, that code flow was using a different mapping than others in the `VersionTagger`. Second commit makes it easier to run _some_ tests. Third commit ... shame on me.

PR #269, opened 2017-01-26
Iron out Python 2.6 incompatibilities for CentOS 6

stevekuznetsov:
Use `.assertTrue()` instead of `.assertIn()`

CentOS 6 only has Python 2.6 available, on which the unit testing
frameworks do not expose a `.assertIn()` helper method. Instead, we can
use `.assertTrue(x in y)`.

---

Only use `rpmbuild --noclean` if it is supported

CentOS 6 only has rpmbuild 4.8 available, on which the `--noclean` option
is not present. We should only pass `--noclean` to `rpmbuild` when we
can determine that it is supported.

---

Explicitly define indicies in formatting statements

Python 2.6 is the latest Python version available on CentOS 6. In Python
2.6, the newer formatting statements must have indexed formatting
specifiers, so you must write

    "{0}{1}".format(foo, bar)

instead of:

    "{}{}".format(foo, bar)

---

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -232,8 +232,13 @@ class BuilderBase(object):
     # Assume that if tito's --no-cleanup option is set, also disable %clean in rpmbuild:
     def _get_clean_option(self):
         if self.no_cleanup:
-            return "--noclean"
-        return "--clean"
+            output = run_command('rpmbuild --help')
+            if '--noclean' in output:
+                return "--noclean"
+            else:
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
@dgoodwin we should see the Travis tests go green on this one. Only things failing are CentOS 6 stuff.

stevekuznetsov:
Cool. That fixes us.

dgoodwin:
sfvocqi

Issue #166, opened 2015-04-16
Traceback when doing 'tito tag --undo' while git repo has uncommitted changes


epienbroek:
I was just trying to use 'tito tag --undo' on a git repo where there are some uncommitted changes. This resulted in a traceback:

```
$ tito tag --undo
Undoing tag: foo-6.1-0.2
Checking for tag [foo-6.1-0.2] in git repo [git@foo:bar.git]


########## ERROR ############
Error running command: git tag -d foo-6.1-0.2 && git reset --merge HEAD^1
Status code: 32768
Command output: Deleted tag 'foo-6.1-0.2' (was ffce22a)
error: Entry 'foo.spec' not uptodate. Cannot merge.
fatal: Could not reset index file to revision 'HEAD^1'.
^[[ATraceback (most recent call last):
  File "/usr/bin/tito", line 23, in <module>
    CLI().main(sys.argv[1:])
  File "/usr/lib/python2.7/site-packages/tito/cli.py", line 225, in main
    return module.main(argv)
  File "/usr/lib/python2.7/site-packages/tito/cli.py", line 682, in main
    return tagger.run(self.options)
  File "/usr/lib/python2.7/site-packages/tito/tagger/main.py", line 101, in run
    self._undo()
  File "/usr/lib/python2.7/site-packages/tito/tagger/main.py", line 152, in _undo
    undo_tag(tag)
  File "/usr/lib/python2.7/site-packages/tito/common.py", line 395, in undo_tag
    output = run_command("git tag -d %s && git reset --merge HEAD^1" % tag)
  File "/usr/lib/python2.7/site-packages/tito/common.py", line 303, in run_command
    raise RunCommandException(command, status, output)
tito.exception.RunCommandException: Error running command: git tag -d foo-6.1-0.2 && git reset --merge HEAD^1
```

FrostyX:
Hello @epienbroek,
thank you for the report. The issue was already fixed

    [jkadlcik@localhost subproject2]$ tito tag --undo
    Creating output directory: /tmp/tito
    Undoing tag: subproject-enum-1.3-5
    Checking for tag [subproject-enum-1.3-5] in git repo [git@github.com:FrostyX/project-for-testing-tito.git]
    ERROR: Cannot undo a tag when the current git state is not clean!

If it still happens to you, feel free to reopen.

PR #270, opened 2017-01-28
Do not undo tags when git state is dirty

stevekuznetsov:
When using `tito tag --undo` in a repository with a dirty
`git` state, no action should be taken.


stevekuznetsov:
Addresses https://github.com/dgoodwin/tito/issues/166. Still need to sign the commit.

dgoodwin:
ack

PR #276, opened 2017-04-01
Switch to using createrepo_c for creating rpm repos and drop end-of-life EL5 stuff

Conan-Kudo:
This PR proposes to change Tito to use [`createrepo_c`](http://rpm-software-management.github.io/createrepo_c/) (a markedly faster and more correct rpm repo generator tool). 

It simultaneously drops the leftover CentOS 5 stuff as it is end-of-life and EPEL for EL5 never offered it.


dgoodwin:
This all looks good to me and tests fine, anyone have objections? ( @xsuchy @jmrodri @stevekuznetsov )

hacking/titotest-fedora-25/Dockerfile
```diff
@@ -1,25 +1,28 @@
...
-           python-mock       \
-           python-nose       \
-           python-blessings  \
-           python-pep8       \
-           python3-mock      \
-           python3-nose      \
-           python3-blessings \
+# Manually install python2 deps since spec won't add them
+RUN dnf -y install                 \
+           'dnf-command(builddep)' \
```
stevekuznetsov: There was a BZ blocking this before, thanks for making this change now.

Conan-Kudo:
@dgoodwin I did not change your HACKING guidelines to bump from Python 2.4 to Python 2.6, as I felt it was out of scope of this PR, but you might want to consider it, post-merge.

xsuchy:
This PR looks good to me. But when I was reviewing this PR I stumbled why createrepo is in BuildRequires in first place. It was added in ea7404d6. And IMHO is not needed at all. It is used during CI when using docker images, but that is not used/need when building *rpm* itself. So it can be safely removed.

Conan-Kudo:
@dgoodwin I'm shipping this change in my Mageia package of tito, since I don't want to depend on the old createrepo (it's slow and does things wrong).

@xsuchy I've added a commit to remove `createrepo_c` from the BRs.

dgoodwin:
Thanks for contributing!

Conan-Kudo:
Thanks for accepting the contribution! :)

PR #277, opened 2017-04-18
VersionTagger support for custom tag format

vrutkovs:
`VersionTagger` creates a long tag - e.g. `project-1.0-3` - but its not required for most projects. This PR adds a new tagger which creates `<version>` tag, which would fit most projects.

TODO:
 * [x] Convert this to a config setting with a string to format
 * [x] Update documentation
 * [x] Add tests

### DIFF
--- a/tito.props.5.asciidoc
+++ b/tito.props.5.asciidoc
@@ -80,6 +80,10 @@ This option is used control the text of git commit message that is used when
 new tag is generated. You can use "%(name)s", "%(release_type)s" and
 "%(version)s" placeholders.
 
+tag_format::
+This option controls the format used in VersionTagger. If not specified
+default format would be '{component}-{version}-{release}'. It won't affect
+other taggers.
 
 KOJI and COPR
 -------------
--- a/src/tito/tagger/main.py
+++ b/src/tito/tagger/main.py
@@ -468,6 +468,7 @@ class VersionTagger(ConfigObject):
                '[%(name)s] %(release_type)s [%(version)s].')
         if self.config.has_option(BUILDCONFIG_SECTION, "tag_commit_message_format"):
             fmt = self.config.get(BUILDCONFIG_SECTION, "tag_commit_message_format")
+        new_version_w_suffix = self._get_suffixed_version(new_version)
         try:
             msg = fmt % {
                 'name': self.project_name,
@@ -482,11 +483,11 @@ class VersionTagger(ConfigObject):
         run_command('git commit -m {0} -m {1} -m {2}'.format(
             quote(msg), quote("Created by command:"), quote(" ".join(sys.argv[:]))))
 
+        new_tag = self._get_new_tag(new_version)
         tag_msg = "Tagging package [%s] version [%s] in directory [%s]." % \
-                (self.project_name, new_version_w_suffix,
+                (self.project_name, new_tag,
                         self.relative_project_dir)
 
-        new_tag = self._get_new_tag(new_version)
         run_command('git tag -m "%s" %s' % (tag_msg, new_tag))
         print
         info_out("Created tag: %s" % new_tag)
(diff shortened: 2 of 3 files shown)


xsuchy:
Please document this tagger in `tito.props.5.asciidoc`.

Personally I would rather see this implemented as new option in TAGCONFIG. So this will get propagated to all other Tagger objects.

dgoodwin:
I think I like @xsuchy's suggestion, would you be willing to move this over to something in the tagconfig section of tito.props @vrutkovs?

vrutkovs:
Agreed, I'm thinking to add a new config setting with a format string, e.g. `{component}-{version}-{release}` - does that sound good to you?

stevekuznetsov:
I think this mostly falls under the conversations @dgoodwin and I had about allowing a user to use a `UserSpecifiedTagger` or something -- just as you said, a format-string with places for the whole NEVRA. I would suggest using the `rpm` field specifiers, so `%{VERSION}`, *etc*.

vrutkovs:
Oh, I like the `UserSpecifiedTagger` name.

I was thinking to use python format syntax for this, but may convert this to rpm-style if needed.

dgoodwin:
I'd lean towards python formatting just for ease of implementation, and because we're kind of distancing ourselves from RPM in some ways, or at least I'd like to. I assume for now we just have name, version, and release supported? (I think those are the only three bits of information tito knows) Should there be any others?

Another possible idea for the name: CustomFormatTagger

vrutkovs:
> I assume for now we just have name, version, and release supported?

At this moment - yes, although its possible to extend this function on request, as it has access to `self`

vrutkovs:
@xsuchy @dgoodwin ping? Tested this and it worked fine for me.

Not sure about the naming, but if you agree `CustomTagger` is better I'll rename it

test/functional/user_specified_tag_tests.py
```diff
@@ -0,0 +1,87 @@
...
+%changelog
+"""
+
+
+def tito(argstring):
+    """ Run Tito from source with given arguments. """
+    return CLI().main(argstring.split(' '))
+
+
+class UserSpecifiedTagTest(unittest.TestCase):
```
dgoodwin: I would request a rename, user specified doesn't seem right here to me as it's not entirely clear what that means, users can always specify which tagger to use. This is more about custom tag formats so I would suggest CustomFormatTagger or FormatTagger, open to other suggestions as well.

test/functional/user_specified_tag_tests.py
```diff
@@ -0,0 +1,87 @@
...
+        out_f.write(contents)
+        out_f.close()
+
+    def create_rpm_package(self):
+        os.chdir(self.full_pkg_dir)
+        self.write_file(os.path.join(self.full_pkg_dir, 'hello_tito.spec'), TEST_SPEC)
+        run_command('git add hello_tito.spec')
+        run_command("git commit -m 'add spec file'")
+
+    def test_tag(self):
```
dgoodwin: Kudos on adding a test, thanks.

tito.props.5.asciidoc
```diff
@@ -80,6 +80,10 @@ This option is used control the text of git commit message that is used when
 new tag is generated. You can use "%(name)s", "%(release_type)s" and
 "%(version)s" placeholders.
 
+tag_format::
+This option controls the format used in UserSpecifiedTagger. If not specified 
+default format would be '{component}-{version}-{release}'. It won't affect 
+other taggers.
```
dgoodwin: Actually looking at this I'm now wondering if this should be part of the base tagger, not a separate implementation. Would this slip into the base VersionTagger, trigger the behavior if the user specified a tag format?

If not, we should move this out of the more global "buildconfig" section and into a "format_tagger" section.

However curious what you think about bringing this into the core tagger itself.

vrutkovs: +1, that would resolve the naming issue

vrutkovs:
Pushed a new commit to merge these functions in `VersionTagger`, updated the PR title

dgoodwin:
Looks like some failing tests in travic.

vrutkovs:
Right, fixed it

dgoodwin:
@vrutkovs thanks for your patience and this contribution, really nice addition IMO.

PR #278, opened 2017-05-04
Custom tag support in tito release

vrutkovs:
A follow-up for https://github.com/dgoodwin/tito/pull/277 - `tito release` should also understand the new tag format


dgoodwin:
LGTM, will merge if no objections today.

Issue #274, opened 2017-03-01
tito tag --use-version option is ignored with ReleaseTagger


rohanpm:
The --use-version option to tito tag is ignored if tito.tagger.ReleaseTagger is used.

tito version:

    $ rpm -q tito
    tito-0.6.10-1.el7.noarch

.tito/tito.props:

  [buildconfig]
  builder = tito.builder.Builder
  tagger = tito.tagger.ReleaseTagger
  changelog_do_not_remove_cherrypick = 0
  changelog_format = %s (%ae)

Steps to reproduce:

- Run `tito tag --use-version 1.2.3`, and save changelog

Expected behavior:

tito produces a tag matching <package-name>-1.2.3-<release>, or raises an error telling me it's invalid to use --use-version and ReleaseTagger together.

Actual behavior:

tito behaves the same as if I'd omitted `--use-version` option, and generates a tag by bumping the current Release.

Additional notes:

It works in tito 0.6.9.

stevekuznetsov:
I'll look into this. Thanks for the bug report.

stevekuznetsov:
@rohanpm could you please provide the RPM version of `tito` where this worked in 0.6.9? The change that immediately looks to be the culprit was already in the tree for the 0.6.9 release, so I just want to verify.

rohanpm:
I don't have RPM version - for earlier versions I tested from git tag, installing to virtualenv via "python setup.py install".

At tag tito-0.6.10-1 the problem was reproducible and at tag tito-0.6.9-1 it wasn't.

I thought that commit 6b57b0bcbcbb869 between 0.6.9 and 0.6.10 looked relevant.

stevekuznetsov:
Yeah, a couple things come to mind -- @dgoodwin we should hash out a test matrix for [`VersionTagger._bump_version()`](https://github.com/dgoodwin/tito/blob/master/src/tito/tagger/main.py#L377-L436) -- somewhere along the line the refactor to that method led to this regression but it's hard to reason about as it is factored today.

lcarva:
I'm also seeing this on `tito-0.6.10-1.fc25.noarch`. Downgrading to `tito-0.6.6-1.fc25.noarch` fixes the issue.

evgeni:
Downgrading to `tito-0.6.9-1.fc25.noarch` is enough.

evgeni:
[...]

PR #283, opened 2017-05-18
force VersionTagger when tito tag is called with --use-version

evgeni:
In 6b57b0b, the code that checked wither we used `--use-version` was removed, because both `VersionTagger` and `ForceVersionTagger` are identical now. However, this breaks setups where the default tagger is `ReleaseTagger` but the user still wants to use `--use-version` sometimes.

Thus let's force `VersionTagger` when `--use-version` was given and only use whatever is in `tito.props` when it was not.

Fixes: #274

Signed-off-by: Evgeni Golov <egolov@redhat.com>

### DIFF
--- a/src/tito/cli.py
+++ b/src/tito/cli.py
@@ -653,8 +653,11 @@ class TagModule(BaseCliModule):
             debug("block_tagging defined in tito.props")
             error_out("Tagging has been disabled in this git branch.")
 
-        tagger_class = get_class_by_name(self.config.get(
-            BUILDCONFIG_SECTION, DEFAULT_TAGGER))
+        if self.options.use_version:
+            tagger_class = get_class_by_name("tito.tagger.VersionTagger")
+        else:
+            tagger_class = get_class_by_name(self.config.get(
+                BUILDCONFIG_SECTION, DEFAULT_TAGGER))
         debug("Using tagger class: %s" % tagger_class)
 
         tagger = tagger_class(config=self.config,


dgoodwin:
Looks correct to me, you good with this @stevekuznetsov ?

evgeni:
oups, typo in the subject, this is `tito`, not [Toto](https://en.wikipedia.org/wiki/Toto_(band)) (fixed)

stevekuznetsov:
The `ReleaseTagger` is calling `self.bump_version()` [here](https://github.com/dgoodwin/tito/blob/master/src/tito/tagger/main.py#L657) and is not overriding `run()` from the `VersionTagger`, so it should be getting `self._use_version` set [here](https://github.com/dgoodwin/tito/blob/master/src/tito/tagger/main.py#L101-L102). So from first glance I would have expected the `ReleaseTagger` to pick up the hard-coded version already. Why doesn't it? Isn't that a bug?

stevekuznetsov:
Ah, sorry -- we pass `release=True` and this forces `self.bump_version()` to not touch the version field, only release. Is this good UX to switch the tagger? We should at the very least emit a warning that we are doing this. In my opinion, though, we should just error in the case where someone is passing `--use-version` with a tagger that can't update version and let the user fix their setup instead of trying to automagically figure out what they meant.

dgoodwin:
I'm a bit nervous about switching out the tagger too (even though that's what we always did). Also in the common ReleaseTagger use case you most likely want to version tag and reset release to 1, which is exactly what VersionTagger does. However it still feels wrong to bypass your custom tagger.

Taggers explicitly supporting manually specifying version (and perhaps a --use-release someday) seems like it is required, and if they don't we would error out. In this case we'd need to modify ReleaseTagger to be able to swap in the new version. 

What do you think @evgeni, and would you be up to implement this?

We also should get a unit test on this situation.

evgeni:
So you say to extend `ReleaseTagger` to actually understand `--use-version`? Yeah, I think I can look into that.

dgoodwin:
Thanks!

evgeni:
Closing in favour of #285.

PR #285, opened 2017-05-22
ReleaseTagger --use-version support

evgeni:
This adds `--use-version` support to `ReleaseTagger` as discussed in #274 and #283.
Basically it just checks if the user passed `--use-version` and calls `_bump_version` with either `release=True` or `release=False` based on that.


dgoodwin:
At some point I'd like to make the "I suppose --use-version" contract a little more clear, maybe with an unimplemented method on base class, but this looks good for now, thanks @evgeni !

PR #286, opened 2017-05-24
use LC_ALL=C.UTF-8 rather than plain C

xsuchy:
addressing:
Python runtime initialized with LC_CTYPE=C (a locale with default ASCII encoding), which may cause Unicode compatibility problems. Using C.UTF-8, C.utf8, or UTF-8 (if available) as alternative Unicode-compatible locales is recommended.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -442,7 +442,7 @@ def run_command_print(command):
     """
     output = []
     env = os.environ.copy()
-    env['LC_ALL'] = 'C'
+    env['LC_ALL'] = 'C.UTF-8'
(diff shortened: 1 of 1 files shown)


dgoodwin:
This appears to have introduced test failures: https://travis-ci.org/dgoodwin/tito/builds/235619754

xsuchy:
The failing test is on CentOS6 right? I cannot reproduce it locally on my F26.

dgoodwin:
hacking/runtests.sh should expose it if it's reproducible and you've got docker configured locally.

PR #287, opened 2017-05-24
Don't append 'None' to Release line with no '%{?dist}' part

pfournier:
When doing --test build, if the spec file 'Release' line contains no '%{?dist}' part, tito will append 'None' at the end of the release option. This change makes sure nothing is added when this part is not present in the line.


pfournier:
Rebased on non-broken master to get a successful travis check.

dgoodwin:
Thanks @pfournier, would it be feasible to add a unit test for this?

pfournier:
Added test to the commit.

PR #300, opened 2017-08-07
python3's map() returns a map object, but we expect sources to be a list

evgeni:


src/tito/builder/main.py
```diff
@@ -614,7 +614,7 @@ def tgz(self):
         debug("Scanning for sources.")
         cmd = "/usr/bin/spectool --list-files '%s' | awk '{print $2}' |xargs -l1 --no-run-if-empty basename " % self.spec_file
         result = run_command(cmd)
-        self.sources = map(lambda x: os.path.join(self.rpmbuild_gitcopy, x), result.split("\n"))
+        self.sources = list(map(lambda x: os.path.join(self.rpmbuild_gitcopy, x), result.split("\n")))
```
stevekuznetsov: Could we maybe use fewer nifty syntax here for clarity?

```python
self.sources = []
for file in result.split("\n"):
    self.sources.append(os.path.join(self.rpmbuild_gitcopy, file))
```

evgeni: Surely could, I was going after the smallest change. But if you want more readable code, that certainly can be done ;)

stevekuznetsov: If you don't mind, I think that would be nice!

evgeni: done in caa7c77

stevekuznetsov:
@evgeni thanks for the contribution! :)

dgoodwin:
Thanks @evgeni! and @stevekuznetsov for merging!

PR #302, opened 2017-11-01
Print mock output when building with MockBuilder

vfreex:
Currently when building RPMs or SRPMs with MockBuilder, tito will not print the
output from mock. This makes diagnosis harder especially in CI
environments.

This PR will do the following changes:

1. If `--quiet` option is not specified, `run_command_print` will be used to
run the `mock` or `rpmbuild` command. Otherwise we will continue to use
`run_command`.
2. `run_command_print` will print error message when an error happens, like what `run_command`
does.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -225,7 +225,8 @@ class BuilderBase(object):
             ' "_binary_filedigest_algorithm md5" %s %s %s --nodeps -bs %s' % (
                 rpmbuild_options, self._get_rpmbuild_dir_options(),
                 define_dist, self.spec_file))
-        output = run_command_print(cmd)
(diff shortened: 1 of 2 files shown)


vfreex:
@stevekuznetsov Thanks for your suggestion. Personally I would prefer to see the output during the build procedure, not at the end of build. That's what `run_command_print` does.

vfreex:
@dgoodwin Any suggestions?

stevekuznetsov:
Hadn't thought about needing to do streamed output -- seems reasonable to me but I'll let @dgoodwin weigh in.

dgoodwin:
Yeah I like this, even for default behaiour, when mock is running I'd rather see what it's doing.

Thanks @vfreex

PR #304, opened 2017-11-01
Fix links in README.md file

mirzal:
Links at the bottom of README.md file are not clickable (raw markdown is displayed instead of HTML). This commit fixes that.

Additionally, typo in "Related Projects" section is fixed


dgoodwin:
Thanks!

PR #307, opened 2017-12-04
Fixing remote_git_name

adammhaile:
remote_git_name now works in releasers.conf as intended.

### DIFF
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -49,7 +49,11 @@ class FedoraGitReleaser(Releaser):
         self.git_branches = \
             self.releaser_config.get(self.target, "branches").split(" ")
 
-        if self.config.has_option(self.target, "remote_git_name"):
+        # check .tito/releasers.conf
+        if self.releaser_config.has_option(self.target, "remote_git_name"):
+            overwrite_checkout = self.releaser_config.get(self.target, "remote_git_name")
+        # fallback to old, incorrect, location: .tito/packages/tito.props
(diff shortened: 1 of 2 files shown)


dgoodwin:
Yeah this looks right, it was documented in a per project tito.props which @adamhaile reports didn't even work, only the global tito.props. 

Regardless the setting is specific for releasers, so it should be in releasers.conf.

Now, are we going to break anyone in this form? Should we perhaps watch to make sure we respect the old location as a fallback? Would you mind adding another clause to use releasers.conf if it's there, otherwise check the old location?

src/tito/release/distgit.py
```diff
@@ -49,6 +49,10 @@ def __init__(self, name=None, tag=None, build_dir=None,
         self.git_branches = \
             self.releaser_config.get(self.target, "branches").split(" ")
 
+        # check .tito/releasers.conf
+        if self.releaser_config.has_option(self.target, "remote_git_name"):
+            overwrite_checkout = self.releaser_config.get(self.target, "remote_git_name")
+        # fallback to old, incorrect, location: .tito/packages/tito.props
         if self.config.has_option(self.target, "remote_git_name"):
```
dgoodwin: Should just be an elif here so the correct location wins out.

Otherwise LGTM!

dgoodwin:
Merging, thanks!

Will appear in next release but yell if you need one sooner.

adammhaile:
Perfect, thank you!

PR #313, opened 2018-01-02
Fix chicken/egg problem with .spec files and includes.

dhilst:
The rpm spec system support including files with the `%include` directive. This enable packagers to achieve some level of code re-usage.

# The chicken/egg problem.

`tito` tries to parse the spec file to get the project name. But when the spec file contains an `%include` directive this parsing fails because the included file can't be found. Usually files are included in conjunction of the `SOURCE` directive. So for example:

```
SOURCE0: %{name}-%{version}.tar.gz
SOURCE1: somecool.macros
...
%include %{SOURCE1}
```

`somecool.macros` is expected to be inside the .src.rpm.

# What this PR does?

This PR add one option `--project-name` to avoid early parsing of the spec file, and search for `SOURCE` lines at the spec files. For each `SOURCE` line found the source file is copied to the .src.rpm. With this is possible to use `%include` directive as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
     license='GPLv2+',
 
     package_dir={
-        'tito': 'src/tito',
+        '': 'src/',
     },
     packages=find_packages('src'),
     include_package_data=True,
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -197,6 +197,21 @@ class BuilderBase(object):
             mkdir_p(d)
         self._check_build_dirs_access(build_dirs)
 
+    def _copy_extra_sources(self):
+        """
+        Copy extra %{SOURCEX} files to the SOURCE folder.
+        """
+        with open(self.spec_file, 'r') as spec:
+            for line in spec.readlines():
+                match = re.match(r'SOURCE[1-9]\d*:(?P<src>.*)', line, re.I)
+                if match is None:
+                    continue
+
+                src = os.path.join(self.rpmbuild_sourcedir, self.tgz_dir, match.group('src').strip())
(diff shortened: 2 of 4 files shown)


dhilst:
I've been doing some testing. The `_copy_extra_sources` can't understand rpm macros. So the above snippet will not work because it will try to copy `%{name}.macros` to source folder which obviously is not want user is expecting :-(

```
Source1: %{name}.macros
```

FrostyX:
Hello @dhilst,
thank you for the contribution and I am truly sorry, that it hung here for so long without merging.

I really like some concepts from this PR, but others not that much. Let me elaborate. The `_copy_extra_sources` is IMHO a really good idea and I have nothing against it. You forgot to call it in `def rpm()` function, but that is a minor issue.

What I am not a fan of is the need to specify `--project-name` and ideally I would like to avoid that.

I would also like to avoid the `p = subprocess.Popen(command, shell=True, ...)` because of the [security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations).

I played with this feature a little bit and I would like to propose another solution (based on this one). I took `_copy_extra_sources` from this PR and fixed the rest by defining a `_sourcedir` for the `get_project_name` function. Please see PR#354. Does it meet your expectations?

FrostyX:
ping @dhilst

FrostyX:
Gently pinging again @dhilst

FrostyX:
Closing in favor of PR#354
If it doesn't fix your issue or there is another use-case that I haven't considered, please let me know.

dhilst:
Sorry I didn't see this, it got lost in a flood of security updates of my node projects 🙄

Yeah Popen shell=True is not cool, at the time I had to automate some builds that already have SOURCE lines so I did the fast and dirty way. I'm happy that it was useful is some way

Issue #312, opened 2018-01-02
Regression with releases in tags


ekohl:
Since 0.6.11 I'm having issues where it creates an incorrect tag. When I tried to tag a package, it created the correct changelog but with an incorrect git tag. https://github.com/theforeman/foreman-packaging/commit/607c00b280cee68f764c8a8055707b8d5928bbee is the commit in question. That created ```rubygem-radcli-1.0.0.fm1_17-1``` but then ```tito release koji-foreman-plugins``` complained it could't find the tag ```rubygem-radcli-1.0.0-1.fm1_17```. Note the release was moved.

A similar regression can be seen with trying to tag https://github.com/theforeman/foreman-packaging/commit/f532f793cb4e385abc040ff428f438aef992d1bf:

```
$ tito tag --keep-version --accept-auto-changelog 
Creating output directory: /tmp/tito
ERROR: Error running command: git log --no-merges --pretty='format:%s (%ae)' --relative nodejs-redux-form-validators-2.1.0.fm1_17-1.fm1_17..HEAD -- .

Status code: 128

Command output: fatal: bad revision 'nodejs-redux-form-validators-2.1.0.fm1_17-1.fm1_17..HEAD'

Traceback (most recent call last):
  File "/usr/bin/tito", line 23, in <module>
    CLI().main(sys.argv[1:])
  File "/usr/lib/python3.6/site-packages/tito/cli.py", line 203, in main
    return module.main(argv)
  File "/usr/lib/python3.6/site-packages/tito/cli.py", line 671, in main
    return tagger.run(self.options)
  File "/usr/lib/python3.6/site-packages/tito/tagger/main.py", line 114, in run
    self._tag_release()
  File "/usr/lib/python3.6/site-packages/tito/tagger/main.py", line 656, in _tag_release
    self._make_changelog()
  File "/usr/lib/python3.6/site-packages/tito/tagger/main.py", line 261, in _make_changelog
    output = self._generate_default_changelog(last_tag)
  File "/usr/lib/python3.6/site-packages/tito/tagger/main.py", line 213, in _generate_default_changelog
    output = run_command(patch_command)
  File "/usr/lib/python3.6/site-packages/tito/common.py", line 426, in run_command
    raise RunCommandException(command, status, output)
tito.exception.RunCommandException: Error running command: git log --no-merges --pretty='format:%s (%ae)' --relative nodejs-redux-form-validators-2.1.0.fm1_17-1.fm1_17..HEAD -- .
```

[...]

PR #314, opened 2018-01-02
WIP: Fix tag_suffix problems in latest release.

dgoodwin:
Introduction of configurable tag formats has caused some issues particularly for users with a tag_suffix defined in their tito.props. 

Issues identified so far:
- tags are now created with name-version-suffix-release, previously this was name-version-release-suffix.
- changelog generation is broken due to how it interprets the data in .tito/packages/pkgname. Here unfortunately we historically stored version-release-suffix. (note: no package name) This implies to reconstruct the last tag we have to assume a prefix of name-, which means we do not *really *have configurable tag formats, at least not where the packagename- prefix is concerned.

So far this PR exposes both problems with tests and corrects them. We still cannot for example, remove or re-order the packagename- prefix in a tag format.

### DIFF
--- a/src/tito/tagger/main.py
+++ b/src/tito/tagger/main.py
@@ -236,6 +236,7 @@ class VersionTagger(ConfigObject):
                 found_changelog = True
 
                 old_version = get_latest_tagged_version(self.project_name)
+                debug("Got old_version: %s" % old_version)
 
                 fd, name = tempfile.mkstemp()
                 write(fd, "# Create your changelog entry below:\n")
@@ -258,6 +259,7 @@ class VersionTagger(ConfigObject):
                 else:
                     if old_version is not None:
                         last_tag = self._get_new_tag(old_version)
+                        debug("last_tag = %s" % last_tag)
                         output = self._generate_default_changelog(last_tag)
                     else:
                         output = self._new_changelog_msg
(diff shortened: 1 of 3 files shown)


dgoodwin:
Working towards a fix for #312.

 CC @vrutkovs

dgoodwin:
I'd really like to transition to package metadata containing the full last tag + path to that package's data, rather than with the packagename- prefix omitted. They're not versioned however.

Pondering something like transitioning to metadata files of: v2 pkgname-0.1-1-mysuffix pkgname/

Existence of 3 tokens here after splitting on empty space implies it's a new v2 metadata file, if there's only 2 we know it's the original v1 and we assume packagename- prefix.

dgoodwin:
Alternatively we could just drop the support for "component-" in the custom tag format, it's assumed, and you cannot control that portion. (which you can't really do today anyhow)

FrostyX:
@dgoodwin,
I know that this PR is two years old, so you probably don't remember the most of it, but any idea of how we should proceed here?

I tried to run tito from this PR and it fixed the traceback described in #312, but I believe the changelog is still generated incorrectly

    Created tag: enum-1.4-1.fm1_17

while the changelog is

    %changelog
    * Thu Mar 12 2020 Jakub Kadlcik <frostyx@email.cz> 1.4-1

Do you want to finish the PR or should I do so? Also, do you remember why the PR is still marked as WIP, is something else necessary to fix the issue? I see you made some suggestions in previous comments, but I am not really sure if it is something that we should do long-term or something that is necessary for closing the issue.

dgoodwin:
Apologies I definitely cannot complete the PR, I am swamped and haven't run tito in like 5 years. :( Feel free to close or complete as you wish.

FrostyX:
> Apologies I definitely cannot complete the PR, I am swamped and haven't run tito in like 5 years. :( Feel free to close or complete as you wish.

Understandable. No problem @dgoodwin, I will try to finish it and consult the result with @ekohl

ekohl:
We've moved away from tagging with tito so this no longer affects the project I'm working on.

FrostyX:
I am trying to finish this PR, and it's almost done, I am just fixing corner cases. Does anyone know if `--use-release` is supposed to be used like this

     tito tag --use-release=123

or like this

    tito tag --use-release="123%{?dist}"

It is AFIAK not documented anywhere, so I am going to do that also.

PR #317, opened 2018-03-02
Avoid double builds with mock

treydock:
According to mock man page the default behavior without a command flag is to run --rebuild.
This change avoids what is essentially doing --rebuild twice.

```
       --rebuild
              If no command is specified, rebuild is assumed. Rebuild the specified SRPM(s). The chroot (including the results directory) is cleaned first, unless --no-clean is specified.

```


dgoodwin:
Thanks!

Issue #316, opened 2018-01-09
build without quiet tries to run cp with a glob, with no plan to expand the glob


aecolley:
Since tito-0.6.11-1.el7.centos.noarch appeared on CentOS 7, "tito build" runs have been failing. A stacktrace is below. The problem is that `_build_in_mock` calls either `run_command_print` or `run_command` depending on the `quiet` flag, but those functions treat their arguments differently.

`run_command` feeds the command to a shell (`commands.getstatusoutput`), which expands glob patterns. However, `run_command_print` merely splits the command on whitespace and feeds it to `subprocess.Popen` with the implied `shell=False` default parameter; therefore there is no shell which might expand the glob.

Proposed fix: change `run_command_print` to use `shell=True` instead of `shlex.split`.

```
Command output: ["cp: cannot stat '/tmp/tmp.EqTGC8kPk3/rpmdump/rpmbuild-murgatroydazCRcx/mockoutput/*.rpm': No such file or directory"]

Traceback (most recent call last):
  File "/usr/bin/tito", line 23, in <module>
    CLI().main(sys.argv[1:])
  File "/usr/lib/python2.7/site-packages/tito/cli.py", line 203, in main
    return module.main(argv)
  File "/usr/lib/python2.7/site-packages/tito/cli.py", line 378, in main
    return builder.run(self.options)
  File "/usr/lib/python2.7/site-packages/tito/builder/main.py", line 156, in run
    self.rpm()
  File "/usr/lib/python2.7/site-packages/tito/builder/main.py", line 1139, in rpm
    self._build_in_mock()
  File "/usr/lib/python2.7/site-packages/tito/builder/main.py", line 1166, in _build_in_mock
    (mock_output_dir, self.rpmbuild_basedir))
  File "/usr/lib/python2.7/site-packages/tito/common.py", line 468, in run_command_print
    raise RunCommandException(command, status, "\n".join(output))
tito.exception.RunCommandException: Error running command: cp -v /tmp/tmp.EqTGC8kPk3/rpmdump/rpmbuild-murgatroydazCRcx/mockoutput/*.rpm /tmp/tmp.EqTGC8kPk3/rpmdump
```

aecolley:
Caused by #316

mtinberg:
[...]

PR #319, opened 2018-03-07
Change way built files are copied to basedir to address issue 316

lukemassa:
https://github.com/dgoodwin/tito/issues/316 points out an issue in which calling out to run_command does not deal with globbing properly. Instead of relying on the shell to do this, we already have a list of rpms, I just cp them one-by-one in the following for loop.


lukemassa:
Just noticed that https://github.com/dgoodwin/tito/pull/318 is addressing essentially the same thing, my mistake, closing this now.

PR #320, opened 2018-04-23
Revert "use built-in shutil.copy2 instead of relying external cp file globbing"

dgoodwin:
Reverts dgoodwin/tito#318

Checking if this will fix the build.


FrostyX:
@dgoodwin do we still want this change?

> Checking if this will fix the build.

Seems like we have no problem building the package now.

PR #324, opened 2018-07-20
Add support for building with Git LFS.

falcon006:


dgoodwin:
Anyone have use for this who can help test? Looks ok to me otherwise and if nobody responds we'll just merge it.

Thanks for submitting @falcon006 .

PR #325, opened 2018-07-25
Add support for SUSE-style detached changelogs

Conan-Kudo:
SUSE-style RPM packaging generally prefers having the RPM changelog detached from the main RPM spec file as a changes file. This file is typically named `<pkg>.changes` (corresponding to `<pkg>.spec`).

At package build time in the Open Build Service, the content of the changes file is reformatted to the RPM changelog structure and appended to the spec file by the build system before `rpmbuild` processes it.

This change adds support for this particular workflow, which is common in communities using an instance of the Open Build Service, such as (open)SUSE.

This was originally authored by @mcalmer. The code was ported to work on Python 3 and adapted for the current tito codebase by @Conan-Kudo.

### DIFF
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -68,7 +68,7 @@ class FedoraGitReleaser(Releaser):
         self.build_targets = build_target_parser.get_build_targets()
 
         # Files we should copy to git during a release:
-        self.copy_extensions = (".spec", ".patch")
+        self.copy_extensions = (".spec", ".changes", ".patch")
 
     def release(self, dry_run=False, no_build=False, scratch=False):
         self.scratch = scratch
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -68,7 +68,7 @@ class FedoraGitReleaser(Releaser):
         self.build_targets = build_target_parser.get_build_targets()
 
         # Files we should copy to git during a release:
-        self.copy_extensions = (".spec", ".changes", ".patch")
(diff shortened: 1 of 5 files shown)


src/tito/cli.py
```diff
@@ -630,6 +630,9 @@ def __init__(self):
                 default=False,
                 help=("Don't automatically create a changelog "
                     "entry for this tag if none is found"))
+        self.parser.add_option("--no-default-changelog", action="store_true",
+                default=False,
+                help=("Don't create the changelog entries but write the header"))
```
dgoodwin: I'm a bit out of touch with my own code, how does this differ from --no-auto-changelog exactly?

juliogonzalez: Ping @mcalmer 

Isn't this related to the fact that we only add headers to the .changes file, but nothing else since we maintain the changelog entries manually and not from the git commits?

Conan-Kudo: If that's the case, then maybe this change isn't necessary, because `--no-auto-changelog` would be sufficient. You're manually managing the changes file entirely at that point...

mcalmer: but does --no-auto-changelog write the header?

Conan-Kudo: No. It writes no entry at all. But why not just invoke `osc vc` and go that route?

Conan-Kudo:
@dgoodwin I think this is in good shape to merge now.

dgoodwin:
Merging, thanks for contributing!

PR #328, opened 2018-12-14
Enable GnuPG signed tags

bcl:
Add 'sign_tag = 1' to the [buildconfig] section of tito.props to enable it.

### DIFF
--- a/src/tito/tagger/main.py
+++ b/src/tito/tagger/main.py
@@ -488,7 +488,13 @@ class VersionTagger(ConfigObject):
                 (self.project_name, new_tag,
                         self.relative_project_dir)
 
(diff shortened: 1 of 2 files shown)


dgoodwin:
Looks good, thanks!

bcl:
Looks like the tests don't like the fallback keyword, but I don't know why. That's valid with py3. Are some of the tests using py2 maybe?

dgoodwin:
Yes quite likely, tito still needs to run on RHEL 7 at the very least and I don't think that's available in the base repos.

bcl:
Rebased onto current master, looks like the tests pass when I run them locally.

dgoodwin:
No bot here @jmrodri :) Merging.

PR #329, opened 2019-01-18
Drop unused fedora-cert Require line from tito.spec

bcl:


FrostyX:
Thank you for the contribution @bcl,
I don't understand how so that this PR doesn't show merge conflicts, but anyway ... it seems like this is already fixed in `e6f4811` (by you).

So I am closing this unmerged.

PR #330, opened 2019-01-18
RPM spec fixes: deps, confusing macro names...

AdamWill:
This patch brings across some fixes which have been made to the
RPM spec in the Fedora package. The %clean section should no
longer be necessary anywhere and is removed. The fedora-cert
dep is no longer needed, and is not available on Fedora Rawhide
(thanks to bcl for that one). The BuildRequire for 'python-devel'
is changed to 'python2-devel'. Finally, two confusing macro
names are changed. The spec was using a custom redefinition of
'python_sitelib', which is very confusing; a Fedora Python dev
saw it and changed it to 'python2_sitelib', assuming it was a
use of the old name for 'python2_sitelib', and that actually
broke Fedora builds. To avoid that kind of confusion, let's
rename our custom '%python_sitelib' to '%our_sitelib', and rename
'%pythonbin' to '%ourpythonbin'.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/tito.spec
+++ b/tito.spec
@@ -1,20 +1,20 @@
 %if 0%{?fedora}
 %global use_python3 1
 %global use_python2 0
-%global pythonbin %{__python3}
-%global python_sitelib %{python3_sitelib}
+%global ourpythonbin %{__python3}
+%global our_sitelib %{python3_sitelib}
 %else
 %global use_python3 0
 %global use_python2 1
 %if 0%{?__python2:1}
-%global pythonbin %{__python2}
-%global python_sitelib %{python2_sitelib}
+%global ourpythonbin %{__python2}
+%global our_sitelib %{python2_sitelib}
 %else
-%global pythonbin %{__python}
-%global python_sitelib %{python_sitelib}
+%global ourpythonbin %{__python}
+%global our_sitelib %{our_sitelib}
 %endif
 %endif
-%{!?python_sitelib: %define python_sitelib %(%{pythonbin} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
+%{!?our_sitelib: %define our_sitelib %(%{ourpythonbin} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
 
 Name: tito
 Version: 0.6.11
@@ -23,8 +23,8 @@ Summary: A tool for managing rpm based git projects
 
 Group: Development/Tools
 License: GPLv2
-URL: http://rm-rf.ca/tito
-Source0: http://rm-rf.ca/files/tito/tito-%{version}.tar.gz
+URL: https://github.com/dgoodwin/tito
+Source0: https://github.com/dgoodwin/tito/archive/tito-%{version}-1.tar.gz
 
 BuildArch: noarch
 %if %{use_python3}
@@ -35,7 +35,7 @@ Requires: python3-bugzilla
 Requires: python3-blessings
 Requires: rpm-python3
 %else
-BuildRequires: python-devel
+BuildRequires: python2-devel
 BuildRequires: python-setuptools
 Requires: python-setuptools
 Requires: python-bugzilla
@@ -64,24 +64,27 @@ BuildRequires: rpm-python3
 Requires: rpm-build
 Requires: rpmlint
 Requires: fedpkg
-Requires: fedora-cert
 Requires: fedora-packager
 Requires: rpmdevtools
-# Cheetah doesn't exist for Python 3, but it's what Mead uses.  We
+# Cheetah used not to exist for Python 3, but it's what Mead uses.  We
 # install it and call via the command line instead of importing the
-# potentially incompatible code
-Requires: python-cheetah
+# previously potentially incompatible code, as we have not yet got
+# around to changing this
+Requires: /usr/bin/cheetah
 
 %description
 Tito is a tool for managing tarballs, rpms, and builds for projects using
 git.
 
 %prep
-%setup -q -n tito-%{version}
-sed -i 1"s|#!.*|#!%{pythonbin}|" bin/tito
+# the weird directory name is because github makes the directory name
+# '(projectname)-(tag)', and the tags for tito have 'tito' in them and
+# '-1' on the end...
+%setup -q -n tito-tito-%{version}-1
+sed -i 1"s|#!.*|#!%{ourpythonbin}|" bin/tito
 
 %build
-%{pythonbin} setup.py build
+%{ourpythonbin} setup.py build
 # convert manages
 a2x -d manpage -f manpage titorc.5.asciidoc
 a2x -d manpage -f manpage tito.8.asciidoc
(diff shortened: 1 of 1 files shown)


AdamWill:
@bcl @dgoodwin

AdamWill:
BTW, on another note, python3-cheetah exists since 3.0.0 which is in all supported Fedoras, so the cheetah stuff could probably be tweaked too. In the meantime I updated this PR to change the requirement to `/usr/bin/cheetah` - requiring `python-cheetah` and assuming it contains the binary is broken for current Fedoras, as the binary is in `python3-cheetah`.

tito.spec
```diff
@@ -1,20 +1,20 @@
...
+%global our_sitelib %{python2_sitelib}
 %else
-%global pythonbin %{__python}
-%global python_sitelib %{python_sitelib}
+%global ourpythonbin %{__python}
+%global our_sitelib %{our_sitelib}
 %endif
 %endif
-%{!?python_sitelib: %define python_sitelib %(%{pythonbin} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
+%{!?our_sitelib: %define our_sitelib %(%{ourpythonbin} -c "from distutils.sysconfig import get_python_lib; print get_python_lib()")}
```
ignatenkobrain: this is not needed since RHEL5 ;)

tito.spec
```diff
@@ -90,8 +90,8 @@ a2x -d manpage -f manpage releasers.conf.5.asciidoc
 
 %install
 rm -rf $RPM_BUILD_ROOT
-%{pythonbin} setup.py install -O1 --skip-build --root $RPM_BUILD_ROOT
-rm -f $RPM_BUILD_ROOT%{python_sitelib}/*egg-info/requires.txt
+%{ourpythonbin} setup.py install -O1 --skip-build --root $RPM_BUILD_ROOT
```
ignatenkobrain: I think this should be %py_install which should be defined to py3_install or to py2_install depending on env

tito.spec
```diff
@@ -90,8 +90,8 @@ a2x -d manpage -f manpage releasers.conf.5.asciidoc
 
 %install
 rm -rf $RPM_BUILD_ROOT
```
ignatenkobrain: this shouldn't be here.

tito.spec
```diff
@@ -90,8 +90,8 @@ a2x -d manpage -f manpage releasers.conf.5.asciidoc
...
-%{pythonbin} setup.py install -O1 --skip-build --root $RPM_BUILD_ROOT
-rm -f $RPM_BUILD_ROOT%{python_sitelib}/*egg-info/requires.txt
+%{ourpythonbin} setup.py install -O1 --skip-build --root $RPM_BUILD_ROOT
+rm -f $RPM_BUILD_ROOT%{our_sitelib}/*egg-info/requires.txt
```
ignatenkobrain: I wonder why this awful line is here.

AdamWill:
Updated: also update URL and Source0 to point to github, as the old site is 404.

tito.spec
```diff
@@ -1,20 +1,20 @@
...
 %if 0%{?__python2:1}
-%global pythonbin %{__python2}
-%global python_sitelib %{python2_sitelib}
+%global ourpythonbin %{__python2}
+%global our_sitelib %{python2_sitelib}
 %else
-%global pythonbin %{__python}
-%global python_sitelib %{python_sitelib}
+%global ourpythonbin %{__python}
+%global our_sitelib %{our_sitelib}
```
Conan-Kudo: All this goop can be dropped for a much simpler structure:

```specfile
%if (0%{?rhel} && 0%{?rhel} < 8)
# RHEL < 8 has no Python 3
%global __python %{__python2}
%else
# Default to Python 3
%global __python %{__python3}
%endif
```

From there, `%python_sitelib`, `%py_build`, `%py_install`, and so on have the desired effect.

tito.spec
Conan-Kudo: Just kill this. :)

tito.spec
```diff
@@ -35,7 +35,7 @@ Requires: python3-bugzilla
 Requires: python3-blessings
 Requires: rpm-python3
 %else
-BuildRequires: python-devel
+BuildRequires: python2-devel
 BuildRequires: python-setuptools
 Requires: python-setuptools
 Requires: python-bugzilla
```
Conan-Kudo: These conditionals can be reworked accordingly to match the suggestion I have up at the top.

tito.spec
```diff
@@ -64,24 +64,27 @@ BuildRequires: rpm-python3
...
 git.
 
 %prep
-%setup -q -n tito-%{version}
-sed -i 1"s|#!.*|#!%{pythonbin}|" bin/tito
+# the weird directory name is because github makes the directory name
+# '(projectname)-(tag)', and the tags for tito have 'tito' in them and
+# '-1' on the end...
+%setup -q -n tito-tito-%{version}-1
+sed -i 1"s|#!.*|#!%{ourpythonbin}|" bin/tito
```
Conan-Kudo: Use `%{__python}` here, per suggestion of how to set the global Python.

tito.spec
```diff
@@ -64,24 +64,27 @@ BuildRequires: rpm-python3
...
 
 %build
-%{pythonbin} setup.py build
+%{ourpythonbin} setup.py build
```
Conan-Kudo: `%py_build`

tito.spec
Conan-Kudo: Indeed, it should be `%py_install`

tito.spec
Conan-Kudo: He's trying to prevent the dependency generator from generating dependencies. But why?

tito.spec
```diff
@@ -100,9 +103,6 @@ cp -a tito.8 %{buildroot}/%{_mandir}/man8/
 # bash completion facilities
 install -Dp -m 0644 share/tito_completion.sh %{buildroot}%{_datadir}/bash-completion/completions/tito
 
-%clean
-rm -rf $RPM_BUILD_ROOT
-
 
 %files
 %doc AUTHORS COPYING
```
Conan-Kudo: License needs to use `%license`. :)

tito.spec
```diff
@@ -114,9 +114,9 @@ rm -rf $RPM_BUILD_ROOT
 %{_bindir}/tito
 %{_bindir}/generate-patches.pl
 %{_datadir}/bash-completion/completions/tito
-%dir %{python_sitelib}/tito
-%{python_sitelib}/tito/*
-%{python_sitelib}/tito-*.egg-info
+%dir %{our_sitelib}/tito
+%{our_sitelib}/tito/*
+%{our_sitelib}/tito-*.egg-info
```
Conan-Kudo: This can be simplified to just the following:

```specfile
%{python_sitelib}/tito/
%{python_sitelib}/tito-*.egg-info
```

tito.spec
```diff
@@ -23,8 +23,8 @@ Summary: A tool for managing rpm based git projects
 
 Group: Development/Tools
 License: GPLv2
-URL: http://rm-rf.ca/tito
-Source0: http://rm-rf.ca/files/tito/tito-%{version}.tar.gz
+URL: https://github.com/dgoodwin/tito
+Source0: https://github.com/dgoodwin/tito/archive/tito-%{version}-1.tar.gz
```
xsuchy: But this is a binary different source than what `tito build --tgz` produce.

AdamWill: Sure, but it has the significant advantage of *actually existing*. The other URL does not (any more). We work with what we can. :)

I did do a quick eyeball compare of the two and they look extremely similar, and the github-generated tarball works fine to actually do a package build. I didn't do an actual diff on the contents, though.

FrostyX:
Hello @AdamWill,
thank you for your contribution. I am sorry that it was not merged for so long.

Meanwhile, I have submitted PR#349 with these exact changes, ... they are even committed under your name because I took them from Fedora distgit. I wasn't aware that there is already this PR opened, sorry. Additionally, I made a few more changes in PR#351.

I believe there are no unmerged changes from this PR, so I am going to close it. Please reopen or fire a new one, if you find out that there is something wrong.

PR #333, opened 2019-03-25
make get_project_name more resilient

evgeni:
before this, packages with a name like "foo-1-setup" would be end up
parsed as "foo" and confused the heck out of the whitelisting code.

~~the new regex is less greedy about numbers by expecting the tag to end
in something that looks like a version (any number of numbers, dots and
hyphens) followed by an optional "dash anything" for the release.~~

split tag instead of parsing it using a regex as nothing says that the version number must start with a number as seen in the gpg-pubkey packages in Fedora:

    gpg-pubkey-facb00b1-570a8081

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -810,7 +810,7 @@ def get_project_name(tag=None, scl=None):
     current working directory. Error out if neither is present.
     """
     if tag is not None:
-        p = re.compile('(.*?)-(\d.*)')
+        p = re.compile('(.*?)-(\d[^-]*)(-[^-]*)?$')
         m = p.match(tag)
         if not m:
             error_out("Unable to determine project name in tag: %s" % tag)
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -810,11 +810,14 @@ def get_project_name(tag=None, scl=None):
     current working directory. Error out if neither is present.
     """
     if tag is not None:
-        p = re.compile('(.*?)-(\d[^-]*)(-[^-]*)?$')
-        m = p.match(tag)
-        if not m:
-            error_out("Unable to determine project name in tag: %s" % tag)
-        return m.group(1)
+        try:
+            (package, _version, _release) = tag.rsplit('-', 2)
+        except ValueError:
+            try:
+                (package, _version) = tag.rsplit('-', 1)
+            except ValueError:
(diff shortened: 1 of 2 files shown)


evgeni:
The CentOS6 error

```
test_simple_build (functional.build_gitannex_tests.GitAnnexBuilderTests) ... ok
ERROR: Unable to determine project name from spec file: /home/sandbox/tito.spec
Try rpm -q --specfile /home/sandbox/tito.spec
Try rpmlint -i /home/sandbox/tito.spec
ERROR
``

also happens on `master`, so this is nothing this PR introduced :)

evgeni:
this now uses `rsplit` instead of an regex, as there are packages whose version does not start with a number (silly, but that's life).

also added tests

FrostyX:
Hello @evgeni, 
I am currently fixing Travis errors in PR#353 and then I will look at this change.

evgeni:
Thanks for fixing the tests!

FrostyX:
Good job @evgeni, thank you for the contribution!
Especially I would like to thank you for putting in the effort and writing also tests.

I tried to refactor the `get_project_name` function a little bit and changed it to 

    if tag is not None:
        return tag.rsplit("-", 2)[0]
    else:
    # Rest of the code unchanged

and all tests pass. Do you think we can have the implementation that simple or did we forget to cover something in tests?

evgeni:
Hey @FrostyX,

I think this highly depends on how forgiving we want to be to the users of that method. And I frankly don't know where it can get it inputs from -- I had one usecase where it failed and my patch fixed it for me, so I didn't look further.

The old code expects at least one dash to be present, followed by one number. And that's also what I tried to implement in my version, keeping in mind that *technically* this method should always see `N-E:V-R` strings, but also accept just `N-V` as that's what the old code accepted. Your version would also accept `N` (and return it correctly). I didn't write tests for the error states tho.

So I think this boils down on how forgiving we want to be for wrong inputs. Your version is certainly easier to read and more forgiving, so if that's acceptable, I'd prefer yours.

FrostyX:
You know what. Let's keep it safe and go with your current implementation. If we want to add some unit tests for a wrong input and refactor that function the future, we surely can.

What do you think?

evgeni:
Sure, works for me :)

evgeni:
Thanks!

PR #334, opened 2019-03-27
tar.py: fix implicit ascii decode of unicode data

sosiouxme:
Given a tarball with a utf8 filename, this was getting errors like `UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 75: ordinal not in range(128)`
The problem is that in order to encode as utf8, it needed to first have a unicode, and so it implicitly decodes to unicode first. However the default decoding is ascii, and the actual contents of the tarball headers contain utf8.
This makes explicit the decoding as utf8, which should cover ascii and utf8 inputs.


sosiouxme:
Nevermind, this rewrote the tarball without incident, but tar didn't like the headers on unpack.

PR #336, opened 2019-04-08
Fix UnicodeDecodeError caused by source tarballs with utf8 names.

awood:


dgoodwin:
Thanks @awood . LGTM, will give folks a little time if anyone wants to comment and then merge.

awood:
Tests are failing for reasons that are unclear to me.  I'll keep working.

awood:
I don't know what's going on here.  The `BuildTitoTests.setUpClass` is failing.  It tries to cd to a temp directory and then untar a tarball of the current version of tito?  But the tarball is screwed up somehow by the UTF8 archive.  I don't think tar.py is corrupting the tar ball because I can run `PYTHONPATH=src python src/tito/tar.py $(date +%s) $(git rev-parse HEAD) test/unit/resources/archivé.tar` and produce a valid tarball.

Ultimately, I don't understand why that functional test is failing and I'm going to need some assistance on figuring out what is going wrong.  I'm currently stumped.

awood:
Just to be clear, I think the proposed patch is doing the right thing.

```
% PYTHONPATH=~/devel/tito/src python src/tito/tar.py 1429725106 3518d720bff20db887b7a5e5dddd411d14dca1f9 test/unit/resources/archivé.tar > test.tar
% PYTHONPATH=~/devel/tito/src python src/tito/tar.py 1429725106 3518d720bff20db887b7a5e5dddd411d14dca1f9 test/unit/resources/archive.tar > test2.tar
% sha256sum test.tar test2.tar test/unit/resources/archive-fixed.tar 
0a946cb69bb5ea3b2acb82724195bf0364c19928c0ca4e3562a327322cb61d18  test.tar
0a946cb69bb5ea3b2acb82724195bf0364c19928c0ca4e3562a327322cb61d18  test2.tar
0a946cb69bb5ea3b2acb82724195bf0364c19928c0ca4e3562a327322cb61d18  test/unit/resources/archive-fixed.tar
```

That's running the proposed code over the `archive.tar` file using known git hashes and timestamps and then comparing them with a reference file.

awood:
Ugg... UTF8 characters in file names are throwing tar.py for a loop.  They're messing up the checksum calculation maybe?  I'll keep digging.  :angry:

awood:
@dgoodwin finally passes.  What a pain in the neck.

dgoodwin:
Not pretty bug, thanks again.

sosiouxme:
:tada: Thanks a bunch guys! :confetti_ball:

PR #340, opened 2019-09-04
Update common.py

adamladd17:
Add in a check to confirm we get the original root commit has in case there are multiple. This can occur when two separate repositories are merged in to one new one. Without this addition the count will always return 0.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -917,6 +917,7 @@ def get_commit_count(tag, commit_id):
         debug("going to use number of commits from initial commit")
         (status, output) = getstatusoutput(
             "git rev-list --max-parents=0 HEAD")
+        output = output.split("\n")[-1]
         if status == 0:
             # output is now inital commit
             (status, output) = getstatusoutput(


FrostyX:
Hello @adamladd17,
thank you for the contribution, and sorry that nobody reviewed it yet.

I think I can merge the PR, it shouldn't cause any issue. But it seems weird that `git rev-list %s..%s --count` returns more than one value. Do you have any repository on which I can test it? We should also write some unit tests covering the function, so I would like to see how a real-life repo looks like.

adamladd17:
Sadly I don't have any available repo to test this on as I only discovered the issue on our offline git server. I don't remember exactly how we got the repo in that state as it was awhile ago but I was able to resolve the problem by just giving the repo a tag. This caused `git describe --match=<tag> <commit_id>` to return successfully. If you would like I can play around and try to recreate the issue with some new empty repos.

FrostyX:
@adamladd17 so, you resolved the issue without requiring this PR? Or is it still needed?
I don't see any way it can break something, so I will gladly merge it ... if it's still relevant.

adamladd17:
I would say it is still relevant for the off chance that people aren't using tags and is worth the merge!

FrostyX:
Okay, thanks for the confirmation. Let's go for it then.

Issue #341, opened 2019-09-16
tito build --test --rpm -i tracebacks on Fedora 31


xsuchy:
On Fedora 31 I got this:
```
~/projects/mock/mock{devel}]130$ tito build --test --rpm -i
....
+ exit 0


Successfully built: /tmp/tito/mock-1.4.19-1.git.8.0c138c8.fc31.src.rpm
        - /tmp/tito/noarch/mock-scm-1.4.19-1.git.8.0c138c8.fc31.noarch.rpm
        - /tmp/tito/noarch/mock-lvm-1.4.19-1.git.8.0c138c8.fc31.noarch.rpm
        - /tmp/tito/noarch/mock-1.4.19-1.git.8.0c138c8.fc31.noarch.rpm
Auto-installing packages:
Traceback (most recent call last):
  File "/usr/bin/tito", line 23, in <module>
    CLI().main(sys.argv[1:])
  File "/usr/lib/python3.7/site-packages/tito/cli.py", line 203, in main
    return module.main(argv)
  File "/usr/lib/python3.7/site-packages/tito/cli.py", line 378, in main
    return builder.run(self.options)
  File "/usr/lib/python3.7/site-packages/tito/builder/main.py", line 157, in run
    self._auto_install()
  File "/usr/lib/python3.7/site-packages/tito/builder/main.py", line 330, in _auto_install
    reinstall = self.package_manager.is_installed(self.project_name, self.build_version)
  File "/usr/lib/python3.7/site-packages/tito/builder/main.py", line 1309, in is_installed
    iv = "%s-%s" % (q.version.decode("utf-8"), q.release.decode("utf-8"))
AttributeError: 'str' object has no attribute 'decode'

$ rpm -q tito
tito-0.6.11-9.fc31.noarch
```

xsuchy:
It is caused by
  https://rpm.org/wiki/Releases/4.15.0
Fix Python 3 to return string data as surrogate-escaped utf-8 strings instead of bytes which nothing else in the API accepts

PR #343, opened 2019-10-02
tito build --test --rpm -i Traceback on Fedora 31

schlupov:
Fixes: #341

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -1340,7 +1340,10 @@ class Rpm(object):
         if not q:
             return False
 
-        iv = "%s-%s" % (q.version.decode("utf-8"), q.release.decode("utf-8"))
+        if isinstance(q.version, bytes):
+            iv = "%s-%s" % (q.version.decode("utf-8"), q.release.decode("utf-8"))
+        else:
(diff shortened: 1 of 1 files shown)


src/tito/builder/main.py
```diff
@@ -1340,7 +1340,12 @@ def is_installed(self, package, version):
         if not q:
             return False
 
-        iv = "%s-%s" % (q.version.decode("utf-8"), q.release.decode("utf-8"))
+        if isinstance(q.version, bytes):
+            iv = "%s-%s" % (q.version.decode("utf-8"), q.release.decode("utf-8"))
```
FrostyX: It looks like this is the only line, that conditionally changes. The `iv_short` definition and `return` lines are the same for both cases. IMHO we can simplify this.

schlupov: Ok, I changed that condition.

FrostyX: This looks much better, thank you.
Also, there seems to be a function `compat.ensure_text`, that might be useful in this situation.

dgoodwin:
Thanks @schlupov

PR #349, opened 2019-12-06
Merge downstream patches

FrostyX:
I found several downstream patches here
https://src.fedoraproject.org/rpms/tito/commits/f31

Without them, tito is not installable (because of `fedora-cert`, but I've applied them all).


FrostyX:
IMHO the Jenkins failure is not caused by this change

    Unable to find image 'titotest-fedora-rawhide:latest' locally

I am merging this PR because there is no negative feedback and those patches are in downstream for some time already.

PR #351, opened 2019-12-20
Fix the Source0 URL and prep phase

FrostyX:
Fixing this build error

    line 40: cd: tito-tito-0.6.12: No such file or directory

We don't need the weird directory naming because it is possible to
pass an optional part of the URL specifying the targeted filename.

See https://fedoraproject.org/wiki/Packaging:SourceURL#Git_Tags

### DIFF
--- a/tito.spec
+++ b/tito.spec
@@ -18,12 +18,16 @@
 
 Name: tito
 Version: 0.6.12
-Release: 1%{?dist}
+Release: 3%{?dist}
 Summary: A tool for managing rpm based git projects
 
 License: GPLv2
 URL: https://github.com/dgoodwin/tito
-Source0: https://github.com/dgoodwin/tito/archive/tito-%{version}-1.tar.gz
+# Sources can be obtained by
+# git clone https://github.com/dgoodwin/tito.git
+# cd tito
+# tito build --tgz
+Source0: %{name}-%{version}.tar.gz
 
 BuildArch: noarch
 %if %{use_python3}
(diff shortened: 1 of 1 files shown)


FrostyX:
I needed this change in order to build the new version in koji.

xsuchy:
You cannot use this, because tito produces binary different tar.gz than `tito build --tgz`. The content is the same, but checksum differs. See:
```
$ tito build --tgz
Checking for tag [tito-0.6.12-1] in git repo [git://github.com/dgoodwin/tito.git]
Building package [tito-0.6.12-1]
Wrote: /tmp/tito/tito-0.6.12.tar.gz
$ wget https://github.com/dgoodwin/tito/archive/fbef7d7/tito-0.6.12.tar.gz
...
$ LC_ALL=C diff tito-0.6.12.tar.gz /tmp/tito/tito-0.6.12.tar.gz
Binary files tito-0.6.12.tar.gz and /tmp/tito/tito-0.6.12.tar.gz differ
```

FrostyX:
Well, in this case, we must set `Source0` to be only a filename, not URL, right? e.g.

    # Sources can be obtained by
    # git clone https://github.com/dgoodwin/tito.git
    # cd tito
    # tito build --tgz
    Source0:    %{name}-%{version}.tar.gz

FrostyX:
Updated, PTAL

FrostyX:
Thank you for review @xsuchy,
I am merging now.

PR #352, opened 2020-01-13
Move to python-blessed

ernestask:
As python-blessings is unmaintained nowadays and is not packaged for
EPEL 8, it might be good to move to an active fork.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1777377

### DIFF
--- a/hacking/titotest-centos-6/Dockerfile
+++ b/hacking/titotest-centos-6/Dockerfile
@@ -25,7 +25,7 @@ RUN yum -y install \
     docbook-style-xsl \
     libxslt \
     rpmdevtools \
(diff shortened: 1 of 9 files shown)


FrostyX:
+1

FrostyX:
The failed check is actually 

    Unable to find image 'titotest-fedora-rawhide:latest' locally
    docker: Error response from daemon: pull access denied for titotest-fedora-rawhide, repository does not exist or may require 'docker login'.

we should fix Travis :-/

FrostyX:
No negative feedback, I am merging.
Thank you, @ernestask!

PR #353, opened 2020-01-30
Fix travis

FrostyX:
Our Travis CI currently fails because:

- We are testing python2 on Fedora rawhide - there is no python2 support for a long time
- `python3-pep8` is retired and replaced with `python3-pycodestyle`
- A lot of failing PEP8 checks, that I currently have no time to fix, so I am temporarily disabling them.

### DIFF
--- a/hacking/runtests.sh
+++ b/hacking/runtests.sh
@@ -63,7 +63,6 @@ readonly default_python2_distros=(
     centos-6
     centos-7
     fedora-27
-    fedora-rawhide
 )
 
 readonly default_python3_distros=(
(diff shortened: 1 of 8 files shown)


FrostyX:
So yeah, Travis tests are finally passing. I wouldn't say that I fixed them because except for fixing the Fedora rawhide Dockerfile, I've actually just temporarily disabled some tests.

PTAL if you want.

evgeni:
Wouldn't it be sufficient to skip the createrepo tests on Rawhide only? They pass fine on el7/f27

FrostyX:
Good point. I will try.

FrostyX:
> Wouldn't it be sufficient to skip the createrepo tests on Rawhide only? They pass fine on el7/f27

Done

evgeni:
The diff looks sensible to me.
Thanks!

FrostyX:
Thank you for the review @evgeni,
I am merging.

Issue #311, opened 2017-12-21
How to use %include %{SOURCEX}?


dhilst:
Hi,

I've used to add some files with handy macros in my specs as:

```
SOURCE1: macros.foo

...

%include %{SOURCE1}
```

Now I'm trying to use tito, but the problem is that it tries to parse the specfile prior the mock chroot creation (I'm using mock builder by the way). So when I run tito I get this error:

![image](https://user-images.githubusercontent.com/392820/34272612-06d3ced2-e678-11e7-8b53-5aa1f9305b5d.png)

What is the right way to handle this? Is using `%include %{SOURCEX}` possible with tito?

I created an verifiable example here: https://github.com/dhilst/hellopy/tree/include

Best regards!

dgoodwin:
I fear this may require some kind of a fix in tito, I don't think this will work today, at least not for a package you're building from source git repo. I'm pretty sure it will work with things like the NoTgzBuilder, but I don't know as I've seen it done here.

https://stackoverflow.com/questions/31411205/rpm-include-a-set-of-constants has an interesting response.

Feels like the rpm sourcedir is going to be tricky in this case.

dhilst:
Thanks for the feed back. I took a look at the SO answer, and already saw that _RPM argh_ article. We have a chicken/egg problem here. Tito tries to parse the spec but spec parsing depends on another file being in BUILDDIR, which is what tito will do afterwards.

I just got curious and take a look at `rpm` source, when it finds the `%include` directive it tries to parse the included file immediately, so we got that error.

I see that tito calls `"rpm -q --qf '%%{name}\n' %s --specfile %s 2> /dev/null | grep -e '^$' -v | head -1` to get the project name. What you think about adding a `--name` or something like that to workaround this case!?

dgoodwin:
Or perhaps allow explicitly setting the package name in tito.props (or the per-package tito.props in a multi-project repo), if set we skip spec parsing for the name. I wonder if there's anywhere else we spec parse early though.

dhilst:
I've added a `--project-name` as proof of concept. This by pass the early parsing error, but failed afterwards inside mock, rpmbuild claims that it couldn't find the included file, and it is right since the sources wasn't exploded yet.

The error:
```
(py3venv) [dhilst@dhilst-thinkpad hellopy]$ cat /var/lib/mock/epel-7-x86_64/result/build.log 
Mock Version: 1.4.7
[...]

PR #354, opened 2020-03-11
Allow including macros through SourceX

FrostyX:
Fix: #311
Supersedes: #313

The `get_project_name` function is called in an early phase when
the temporary `SOURCES` directory for a current build doesn't exist yet.
Therefore as a little workaround, we set `_sourcedir` to an upstream
directory containing the package spec file.

Defining a custom `_sourcedir` (and therefore not using
`~/rpmbuild/SOURCES`) shouldn't be a problem because tito builds
should not depend on anything outside of the upstream git repository.

The use-case described in PR#313 works as expected.

        SOURCE1: somecool.macros
        %include %{SOURCE1}

For multi-spec projects, where you have e.g

        .
        ├── foo
        │   ├── foo.spec
        ├── bar
        │   ├── bar.spec

It is possible to include the common macros from a parent (or some
other) directory by having a symlink.

        cd foo
        ln -s ../somecool.macros ./

Or possibly some other way which I am currently not aware of
because I wasn't able to e.g. persuade `%include` to include from
the parent directory.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -197,6 +197,21 @@ class BuilderBase(object):
             mkdir_p(d)
         self._check_build_dirs_access(build_dirs)
 
+    def _copy_extra_sources(self):
+        """
+        Copy extra %{SOURCEX} files to the SOURCE folder.
+        """
+        with open(self.spec_file, 'r') as spec:
+            for line in spec.readlines():
+                match = re.match(r'SOURCE[1-9]\d*:(?P<src>.*)', line, re.I)
+                if match is None:
+                    continue
+
+                src = os.path.join(self.rpmbuild_sourcedir, self.tgz_dir, match.group('src').strip())
+                debug("Copying %s -> %s" % (src, self.rpmbuild_sourcedir))
+                shutil.copy(src, self.rpmbuild_sourcedir)
+
+
     def srpm(self, dist=None):
         """
         Build a source RPM.
@@ -207,6 +222,8 @@ class BuilderBase(object):
 
         if self.test:
             self._setup_test_specfile()
+        
+        self._copy_extra_sources()
 
         debug("Creating srpm from spec file: %s" % self.spec_file)
         define_dist = ""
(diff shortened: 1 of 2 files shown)


FrostyX:
@xsuchy can you please briefly look at this PR to check that I didn't introduce any security issue?

src/tito/builder/main.py
```diff
@@ -197,6 +197,20 @@ def _create_build_dirs(self):
             mkdir_p(d)
         self._check_build_dirs_access(build_dirs)
 
+    def _copy_extra_sources(self):
+        """
+        Copy extra %{SOURCEX} files to the SOURCE folder.
+        """
+        with open(self.spec_file, 'r') as spec:
+            for line in spec.readlines():
```
xsuchy: I am afraid this will not work when there are macros in the file name. I would rather execute `spectool -S self.spec_file` and parse this output.

xsuchy:
Otherwise LGTM

src/tito/builder/main.py
FrostyX: Good call, thank you. Updated.

ehelms:
This breaks the FetchBuilder as it makes the presumption that tgz_dir is defined for all builders inheriting BaseBuilder without defining a sane default:

```
      File "/usr/bin/tito", line 11, in <module>
        load_entry_point('tito==0.6.13', 'console_scripts', 'tito')()
      File "/usr/lib/python3.7/site-packages/tito/cli.py", line 895, in main
        CLI().main(sys.argv[1:])
      File "/usr/lib/python3.7/site-packages/tito/cli.py", line 203, in main
        return module.main(argv)
      File "/usr/lib/python3.7/site-packages/tito/cli.py", line 598, in main
        scratch=self.options.scratch)
      File "/usr/lib/python3.7/site-packages/tito/release/main.py", line 463, in release
        self._koji_release()
      File "/usr/lib/python3.7/site-packages/tito/release/main.py", line 528, in _koji_release
        builder.srpm(dist=disttag)
      File "/usr/lib/python3.7/site-packages/tito/builder/main.py", line 227, in srpm
        self._copy_extra_sources()
      File "/usr/lib/python3.7/site-packages/tito/builder/main.py", line 209, in _copy_extra_sources
        src = os.path.join(self.rpmbuild_sourcedir, self.tgz_dir, source)
    AttributeError: 'FetchBuilder' object has no attribute 'tgz_dir'
```

ehelms:
Elevated to an issue here -- https://github.com/rpm-software-management/tito/issues/367

PR #356, opened 2020-03-25
Fix travis CI tests for Fedora Rawhide

FrostyX:
Do not run external ls -lR command to debug directory content
    
When running tests for fedora rawhide in Travis, this debug line
fails with following error:
    
    Command output: ls: cannot access
    '/tmp/tmpjw15wzfs-titotest/rpmbuild-titotestpkg24v2hy67/SOURCES/titotestpkg-0.0.1/':
    Operation not permitted
    
Even when changing the debug line to just
    
    debug('', 'ls -l /')
    
it still fails. Let's just use `os.listdir`

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -536,8 +536,8 @@ class Builder(ConfigObject, BuilderBase):
         run_command("cd %s/ && tar xzf %s" % (self.rpmbuild_sourcedir,
             self.tgz_filename))
 
-        # Show contents of the directory structure we just extracted.
-        debug('', 'ls -lR %s/' % self.rpmbuild_gitcopy)
+        debug("Show contents of the directory structure we just extracted:\n%s"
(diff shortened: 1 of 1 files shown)


FrostyX:
Travis tests are passing now.

ernestask:
Huh. Any idea what `ls` chokes on that `os.listdir()` eats up?

Unrelated note: have you considered GitHub Actions? It’s free for public repositories.

FrostyX:
> Huh. Any idea what ls chokes on that os.listdir() eats up?

Only if I'd know.

> Unrelated note: have you considered GitHub Actions? It’s free for public repositories.

File an issue please and let's look on that.

FrostyX:
I am going to merge this, so we don't have Travis failing for all of our new pull requests. I know I didn't leave it here long enough for reviews, so in case something is wrong here, please ping me and we will deal with it in a follow-up PR.

PR #361, opened 2020-03-26
Describe how to release tito

FrostyX:
So we can be sure that every maintainer does it the same way.

### DIFF
--- a/.tito/releasers.conf
+++ b/.tito/releasers.conf
@@ -32,7 +32,6 @@ git_url = git://example.com
 releaser = tito.release.FedoraGitReleaser
 branches = master el6 epel7 epel8 f30 f31 f32
 
-[copr-dgoodwin]
+[copr]
 releaser = tito.release.CoprReleaser
-project_name = tito
-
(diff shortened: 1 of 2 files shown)


FrostyX:
I used this in Copr releaser

    project_name = @rpm-software-management/tito

I suppose that if we transfer the project to the rpm-software-management namespace on GitHub, we probably can have also Copr project with this name. But let's wait with merging until its done.

ernestask:
Or, if dogfooding tito, with the same GitHub Actions (I’ll get to writing up a proposal eventually). Otherwise, I don’t see anything to complain about.

FrostyX:
> Or, if dogfooding tito, with the same GitHub Actions (I’ll get to writing up a proposal eventually).

I would prefer that. I think that when implementing these RFEs
https://github.com/dgoodwin/tito/milestone/2
the release process will be also quite simplified

FrostyX:
We already got the Copr project, so this is not blocked anymore
https://copr.fedorainfracloud.org/coprs/g/rpm-software-management/tito/

PR #368, opened 2020-04-18
Use mkpath instead of running mkdir -p command

### LINKED ISSUE #338: Tito init fails when space exists within path to CWD
If there is a space in the CWD, i.e.:

```
> pwd
/mnt/c/Users/Joe Schmo/IdeaProjects/mynewproject
```

Then `tito init` fails with the error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/c/Users/Joe Schmo/IdeaProjects/mynewproject/.tito/tito.props'
```

In reviewing the resulting filesystem the following empty directory was created in the CWD:
`Schmo/IdeaProjects/mynewproject/.tito/`,
i.e.
`/mnt/c/Users/Joe Schmo/IdeaProjects/mynewproject/Schmo/IdeaProjects/mynewproject/.tito/`

```
$ rpm -q tito
tito-0.6.11-8.fc30.noarch
```

FrostyX:
Fix #338

We want to use as few bash commands instead of python functions as
possible anyway, but this rewrite fixes #338

### DIFF
--- a/src/tito/cli.py
+++ b/src/tito/cli.py
@@ -697,7 +697,7 @@ class InitModule(BaseCliModule):
         propsfile = os.path.join(rel_eng_dir, TITO_PROPS)
         if not os.path.exists(propsfile):
             if not os.path.exists(rel_eng_dir):
-                getoutput("mkdir -p %s" % rel_eng_dir)
+                os.makedirs(rel_eng_dir)
                 print("   - created %s" % rel_eng_dir)
 
             # write out tito.props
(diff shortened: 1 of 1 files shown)


src/tito/cli.py
```diff
@@ -697,7 +698,7 @@ def main(self, argv):
         propsfile = os.path.join(rel_eng_dir, TITO_PROPS)
         if not os.path.exists(propsfile):
             if not os.path.exists(rel_eng_dir):
-                getoutput("mkdir -p %s" % rel_eng_dir)
+                mkpath(rel_eng_dir)
```
Conan-Kudo, line 701: Why not use [`os.makedirs()`](https://docs.python.org/3/library/os.html#os.makedirs)?

FrostyX: True, what was I thinking. Fixed.
Thank you @Conan-Kudo

FrostyX:
Thank you for the review

PR #372, opened 2020-04-22
Add a possibility to have full datetime entries in changelog

### LINKED ISSUE #305: Support for full timestamps for RPM changelogs
Currently, tito only supports the regular changelog date format (like `Wed Nov  1 2017`). However, since RPM 4.14, the full date structure is now supported (like `Wed Nov  1 11:08:13 EDT 2017`) for changelog entries.

Can a setting be added to support this for changelogs?

FrostyX:
Fix #305

### DIFF
--- a/src/tito/tagger/main.py
+++ b/src/tito/tagger/main.py
@@ -68,7 +68,7 @@ class VersionTagger(ConfigObject):
                 self.spec_file_name)
         self.keep_version = keep_version
 
-        self.today = strftime("%a %b %d %Y")
+        self.today = self._changelog_date()
         (self.git_user, self.git_email) = self._get_git_user_info()
         git_email = self.git_email
         if git_email is None:
@@ -162,6 +162,12 @@ class VersionTagger(ConfigObject):
         # Everything looks good:
         print
         undo_tag(tag)
+
+    def _changelog_date(self):
+        option = (BUILDCONFIG_SECTION, "changelog_date_with_time")
+        if self.config.has_option(*option) and self.config.getboolean(*option):
+            return strftime("%a %b %d %T %Z %Y")
+        return strftime("%a %b %d %Y")
 
     def _changelog_remove_cherrypick(self, line):
         """
--- a/test/unit/tagger_tests.py
+++ b/test/unit/tagger_tests.py
@@ -1,0 +1,38 @@
+import unittest
+import mock
+from datetime import datetime
+from tito.tagger import VersionTagger
+from tito.compat import PY2, RawConfigParser
+
+from unit import is_epel6
+if is_epel6:
+    import unittest2 as unittest
+else:
+    import unittest
+
+
+def strftime_mock(s):
+    if not PY2:
+        from datetime import timezone
+    tzinfo = None if PY2 else timezone.utc
+    someday = datetime(2020, 4, 22, 15, 2, tzinfo=tzinfo)
+    return someday.strftime(s)
+
+
+class TestVersionTagger(unittest.TestCase):
+    def setUp(self):
+        self.config = RawConfigParser()
+
+    @mock.patch("tito.tagger.main.strftime", strftime_mock)
+    def test_today(self):
+        tagger = VersionTagger(self.config)
+        assert tagger.today == "Wed Apr 22 2020"
+
+    @unittest.skipIf(PY2, "It is not trivial to represent timezone in python2 because"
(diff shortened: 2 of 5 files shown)


FrostyX:
In `tito.props` we parse two sections - `[buildconfig]` and `[tagconfig]`. It would make sense to add this option to `tagconfig` but it seems like we abandoned it because we have other options, that are clearly tag-related and they are in `buildconfig` (e.g. `tag_suffix`, `tag_format`, `changelog_format, ...)

Anyone knows a story behind this?

xsuchy:
It makes more sense to me in tagconfig. Otherwise LGTM.

tito.props.5.asciidoc
```diff
@@ -69,6 +69,12 @@ changelog_do_not_remove_cherrypick::
 If set to 0, it will not remove from cherry picked commits the part "(cherry
 picked from commit ...)"
 
+changelog_datetime::
```
Conan-Kudo, line 72: `changelog_date_with_timestamp` is probably a better name for this option...

FrostyX: A timestamp is usually an integer (number of seconds from 1970). This name would be confusing because we don't want to put that timestamp into the changelog. Maybe rather `changelog_date_with_time`?

FrostyX:
> It makes more sense to me in tagconfig. Otherwise LGTM.

Me too, but there are currently only two options in `tagconfig` (added in 2013 by yourself) and the rest of them are in `buildconfig`. Seems like we abandoned the `tagconfig` section since then and put everything into `buildconfig`. See my previous comment.

I was thinking, that in the future, maybe we could throw away these two config sections and merge them into `[main]` or something.

tito.props.5.asciidoc
Conan-Kudo: Works for me!

FrostyX: Updated

tito.props.5.asciidoc
```diff
@@ -69,6 +69,12 @@ changelog_do_not_remove_cherrypick::
 If set to 0, it will not remove from cherry picked commits the part "(cherry
 picked from commit ...)"
 
+changelog_date_with_time::
+By default, changelog dates don't contain time (e.g. Wed Nov 1 2017). When
+this option is set to `True`, changelog entries are generated in a datetime
+format (e.g. Wed Nov 1 11:08:13 EDT 2017). This feature requires RPM 4.14,
+therefore it is supported only on Fedora 27, EPEL7 and higher.
```
Conan-Kudo, line 76: Fedora 27, RHEL/CentOS 8, SLE/openSUSE Leap 15, and Mageia 7 are the minimum version distros where this feature works.

FrostyX: Updated

src/tito/tagger/main.py
```diff
@@ -163,6 +163,12 @@ def _undo(self):
         print
         undo_tag(tag)
 
+    def _changelog_date(self):
+        option = (BUILDCONFIG_SECTION, "changelog_date_with_time")
+        if self.config.has_option(*option) and self.config.getboolean(*option):
+            return strftime("%a %b %d %H:%M:%S %Z %Y")
```
Conan-Kudo, line 169: This is the correct date format string: `%a %b %d %T %Z %Y`

Conan-Kudo: Though your string works too. 👍

FrostyX: I've updated it and used `%T`, thanks.

test/unit/tagger_tests.py
```diff
@@ -0,0 +1,38 @@
...
+        self.config = RawConfigParser()
+
+    @mock.patch("tito.tagger.main.strftime", strftime_mock)
+    def test_today(self):
+        tagger = VersionTagger(self.config)
+        assert tagger.today == "Wed Apr 22 2020"
+
+    @unittest.skipIf(PY2, "It is not trivial to represent timezone in python2 because"
+                          "datetime.timezone is python3 only and pytz is not in EPEL. This feature"
+                          "is for F27+ and EPEL7+ so we don't need to test it for python2")
```
Conan-Kudo, line 33: EPEL8+

FrostyX: Updated

FrostyX:
Thank you for the review guys.

PR #373, opened 2020-04-24
Move bugzilla code to a separate file

### LINKED ISSUE #358: Make bugzilla dependency optional
The bugzilla module seems an artifact of the Fedora ecosystem, not needed if I'm not building for RedHat/Fedora/EPEL...

```
[~/relup/tito/tito(ERR) (master)] ➔ tito build --tgz
  File "/usr/bin/tito", line 19, in <module>
    from tito.cli import CLI
  File "/usr/share/python3.8/site-packages/tito/cli.py", line 23, in <module>
    from tito.common import find_git_root, error_out, debug, get_class_by_name, \
  File "/usr/share/python3.8/site-packages/tito/common.py", line 33, in <module>
    from bugzilla.rhbugzilla import RHBugzilla
ModuleNotFoundError: No module named 'bugzilla'
[~/relup/tito/tito(ERR) (master)] ➔
```

FrostyX:
Fix #358

Previously the bugzilla-specific code was implemented in `tito.common`
which is imported from everywhere. Therefore missing dependency
to `python3-bugzilla` resulted in failure even for simple commands
such as `tito build --tgz`.


FrostyX:
@glensc, can you please try if this fixes your issue?

It fixes the example with `tito build --tgz` as you requested but I don't know if it affects you elsewhere. Do you require to make the dependency to be optional even in spec file? Does it fail in some other use-case of yours? I think it should fail only when attempting to use `FedoraGitRelease` and `ObsReleaser`.

FrostyX:
I am merging. If it fixes only a part of your use-cases, please re-open the issue.

PR #374, opened 2020-04-26
Make yes or no input less aggressive

### LINKED ISSUE #364: make prompt less aggressive
Today I was working on release.
I accidentaly wrote `yy` instead of `y` and whole release was killed.
I guess it would be better to get user a chance to retry? 

```python
Do you wish to proceed with commit? [y/n] yy
Fine, you're on your own!
  File "/usr/lib/python3.7/site-packages/tito/cli.py", line 598, in main
    scratch=self.options.scratch)
  File "/usr/lib/python3.7/site-packages/tito/release/distgit.py", line 77, in release
    self._git_release()
  File "/usr/lib/python3.7/site-packages/tito/release/distgit.py", line 101, in _git_release
    self._git_user_confirm_commit(self.package_workdir)
  File "/usr/lib/python3.7/site-packages/tito/release/distgit.py", line 199, in _git_user_confirm_commit
    sys.exit(1)
SystemExit: 1
```

FrostyX:
Fix #364

Previously all non-yes input was considered to be "no". Which is not
what we typically want. There should be expected "yes" inputs, expected
"no" inputs and everything else should be considered to be an uknown
input and should be treated this way.

I think that printing an error for unknown input is too verbose and have
no added value. Simply repeating the question is the way to go here:

    ##### Please review the above diff #####
    Do you wish to proceed with commit? [y/n] foo
    Do you wish to proceed with commit? [y/n] y
    Proceeding with commit.

### DIFF
--- a/src/tito/release/main.py
+++ b/src/tito/release/main.py
@@ -102,12 +102,16 @@ class Releaser(ConfigObject):
     def _ask_yes_no(self, prompt="Y/N? ", default_auto_answer=True):
         if self.auto_accept:
             return default_auto_answer
-        else:
-            if PY2:
-                answer = raw_input(prompt)
-            else:
-                answer = input(prompt)
-            return answer.lower() in ['y', 'yes', 'ok', 'sure']
+
+        yes = ['y', 'yes', 'ok', 'sure']
+        no = ['n', 'no', 'nah', 'nope']
+        answers = yes + no
+
+        while True:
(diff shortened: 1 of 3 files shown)


FrostyX:
Hello @StykMartin, do you want to check if this fixes the issue for you?

FrostyX:
Thank you for the review @StykMartin

PR #378, opened 2020-05-28
When extra source file exist, then do not copy it

jirihnidek:
* Extra source file can be copied/moved by some other tool
* When the source file already exists in destination path, then
  do not try to copy it once again
* This issue was observed, when we tried to build
  subscription-manager on Fedora 32 using tito


FrostyX:
Hello, thank you for the PR @jirihnidek!
Looks good to me.

jirihnidek:
Hi @FrostyX,
Thanks for merging into master.

PR #384, opened 2020-07-23
Add Docker support

devodev:
**What:**
This PR adds a Dockerfile as well as a docker-entrypoint.sh script to build RPM's with tito on CentOS8.
Documentation has also been added in README.md under the 'Docker' section.

**Why:**
Tito is already providing a way to build RPM packages reliably. This is extending it further!
We have been using something similar internally with great success and I thought it might be helpful to the community.

**Caveats:**
The current Dockerfile relies solely on official CentOS repositories and does not provide a way to add/enable repositories. IMO, it would be fairly easy to add args/environment variables to pass repositories at build or runtime depending on user needs. For us, we were baking our private repositories inside the Dockerfile to leverage layer caches.

**More:**
This is tested with the tito repository itself and confirmed working.

The PR will probably need refining and maybe need to be transitioned to it's own directory if we would want to adapt it to other distros and so on. Same thing with the documentation, which is only 'functional' to say the least.

P.S.: Sorry about the first try, I used the wrong author..

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,4 +12,4 @@ Miroslav Suchý <msuchy@redhat.com>
 
 and: Adam Miller, Alex Wood, Aron Parsons, Brenton Leanhardt, Ivan Nečas, John Eckersberg,
 Kenny MacDermid, Lukáš Zapletal, Luke Meyer, Marian Csontos, Martin Bačovský, Michael Stead,
-Mike McCune, mscherer, Paul Morgan, Sean P. Kane, Steve 'Ashcrow' Milner
+Mike McCune, mscherer, Paul Morgan, Sean P. Kane, Steve 'Ashcrow' Milner, Alexandre 'devodev' Barone
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,60 @@
+ARG CENTOS_VERSION="8.2.2004"
+FROM centos:${CENTOS_VERSION}
+
+# This is used in the entrypoint to set
+# the builder git config
+ENV GIT_USERNAME "user"
+ENV GIT_EMAIL    "user@example.com"
+
+# Update cache and install tito dependencies,
+# as well as the 'Development Tools' package group
+RUN dnf clean all \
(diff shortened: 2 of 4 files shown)


FrostyX:
Hello @devodev,
thank you for the PR, it looks quite cool, but ...

Maybe I am missing something, in that case, please correct me but to me, it seems that we are trying to re-invent what [Mock](https://github.com/rpm-software-management/mock) does. 

If the purpose is to (1) somewhere have a git repo maintained with Tito, and (2) have a client machine with Fedora, and then being able to build that package for CentOS 8 using that host, then I would go with

    $ git clone https://github.com/FrostyX/tracer.git  # Or some other project maintained with Tito
    $ cd tracer
    $ tito build --srpm --test
    ... 
    Wrote: /tmp/tito/tracer-0.7.4-1.git.1.a0eb36d.fc32.src.rpm
    $ mock -r epel-8-x86_64 /tmp/tito/tracer-0.7.4-1.git.1.a0eb36d.fc32.src.rpm

Am I missing something? Is there some fundamental difference between this and the proposed PR?

devodev:
Hello @FrostyX,

You are right that this is very similar to Mock indeed. It is more of an alternative to it and now that you have mentioned it, I think this would make more sense in its own repository, as a standalone tool.

Thank you for the feedback, I will close this PR.

devodev

FrostyX:
I just realized that I forgot to mention one more important thing.
Mock can now bootstrap its buildroot from a Docker image, please see https://github.com/rpm-software-management/mock/wiki/Release-Notes-1.4.20#container-image-for-bootstrap it's really interesting.

> It is more of an alternative to it and now that you have mentioned it, I think this would make more sense in its own repository, as a standalone tool.

+1, sounds good to me. A lot of users use Mock in complement to Tito, so I would like to prevent duplicating its features here.
It is a cool alternative though.

devodev:
Thank you for that, it looks very interesting. I appreciate you taling the time.

devodev

PR #385, opened 2020-07-25
Use --no-rebuild-srpm for scratch builds in KojiReleaser

### LINKED ISSUE #382: Can KojiReleaser with builder.test=1 use Koji's new (1.21) --no-rebuild-srpm option?
Koji-1.21 now [allows skipping SRPM rebuild for scratch builds](https://pagure.io/koji/pull-request/2083).  This seems perfect to me for doing quicker test builds.  I'm not sure if there would be any downsides.

FrostyX:
Fix #382

### DIFF
--- a/src/tito/release/main.py
+++ b/src/tito/release/main.py
@@ -484,8 +484,12 @@ class KojiReleaser(Releaser):
         if 'KOJI_OPTIONS' in self.builder.user_config:
             koji_opts = self.builder.user_config['KOJI_OPTIONS']
 
-        if self.scratch or ('SCRATCH' in os.environ and os.environ['SCRATCH'] == '1'):
+        scratch = self.scratch or ('SCRATCH' in os.environ and os.environ['SCRATCH'] == '1')
+        if scratch:
             koji_opts = ' '.join([koji_opts, '--scratch'])
+
+        if scratch and (self.test or self.builder.test):
+            koji_opts = ' '.join([koji_opts, '--no-rebuild-srpm'])
 
         if self.profile:
             koji_opts = ' '.join(['--profile', self.profile, koji_opts])


FrostyX:
Thank you for the review @jflorian,
I updated the condition to accommodate also the `builder.test = 1` setting from the `releasers.conf` config file.

PTAL

FrostyX:
By the way, please nevermind the failing Travis CI, it is not related to this change. I tried to resubmit a build from `master` that previously succeeded and it fails now, ... I will fix it in a separate PR.

jflorian:
@FrostyX that now works, but I discovered that Koji fails if passed `--no-rebuild-srpm` without also being passed `--scratch`.  That doesn't seem correct to me and I'm wondering if you concur and that I should raise this issue with the  Koji project.   It seems like an unnecessary limitation but in reviewing the PR [Add rebuild_srpm option of BuildTask](https://pagure.io/koji/issue/1719), I'm inclined to think it will go nowhere.

I can appreciate their reasoning.  If I'm not mistaken, the concept of a "test" build is entirely with tito and while similar to Koji's scratch option, brings something different.  I've also just learned I can set a tag "extra" option `rebuild_srpm=False` and maybe that's a better approach for me, though it would mean about twice as much tag maintenance for each distro rel branch.

FrostyX:
> If I'm not mistaken, the concept of a "test" build is entirely with tito and while similar to Koji's scratch option

Yep, "test" in Tito terms means building from the latest commit instead of the latest tag. Scratch builds are koji term for builds that won't end up in a release.

> but I discovered that Koji fails if passed --no-rebuild-srpm without also being passed --scratch. That doesn't seem correct to me and I'm wondering if you concur and that I should raise this issue with the Koji project. 

In the #382 you linked https://pagure.io/koji/pull-request/2083 that explicitly mentions skipping the SRPM rebuild for scratch builds so it feels like the koji `--no-rebuild-srpm` should be used only if `--scratch` was passed to Tito. I am gonna check with Koji guys and write here some summary.

tkopecek:
As @FrostyX write - Yes, it is allowed only for scratch builds. For production ones user can't force this option.

FrostyX:
Thank you @tkopecek, so I've updated the patch to only use the `--no-rebuild-srpm` koji parameter only when both `--scratch` and `--test` (or their config options) are used for Tito.

FrostyX:
@jflorian what do you think about the PR in this state?

jflorian:
@FrostyX that looks fine.  It doesn't help my use case, but perhaps my practices should change.  I've done almost all of my test builds without `--scratch` because I want those test builds to land in a dist repo for consumption by test clients.  If I use `--scratch` then those clients must install rpms using Koji URLs which can be a pain.  I understand that if the Koji project wants to retain this hard line, then there's nothing tito can do about it.  However, for what most users would want and expect, what you have now looks correct.

tkopecek:
Btw, isn't fedora infra using somwhere scratch repos /every scratch build trigger a simple repository with that build/?

FrostyX:
@jflorian, Okay, thank you for the review. I am merging.

Issue #377, opened 2020-04-27
Allow tagging pre-release versions with `~` tilde character


praiskup:
RPM and Deb distributions for some time allow to use `~suffix` in software
pre-release versions (or distro releases).  The benefit is that tagged
`Version: 2~dev` is _obviously_ devel  only version, but at the same time
users can rely on the fact that `2~dev < 2`, in contrast to `2_dev > 2`.

So maintainers can first bump to `2~dev` and later "bump" to `2` only --
and still there's guaranteed upgrade path from pre-release to production
release.

It would be nice if tito allowed us to do

  `tito tag --use-version=something-with-tilde`

The problem here is that git doesn't allow us to create tags with `~` in
name, that's because `~` has special meaning in get version jargon.  So we
would have to have some clever `version to tag` and `tag to version`
translation.  E.g.

  `tito tag --use-version=2~dev1`

Would create `<name>-<version>_prerelease_dev1-<release>` tag?  Ideas?
Or `s/~/@/`?

Conan-Kudo:
Using `@` would be bad, since that has meaning in software ecosystems to be equivalent to `=`. I think a magic phrase would probably be better...

xsuchy:
And
```
$ rpmdev-vercmp 1@1 1
1@git123 > 1
```
So proper tagged build will not upgrade test version. :(

PR #386, opened 2020-09-11
substitute tilde ~ with at @ for git tags

Blaimi:
Closes: #377

This is probably not the cleanest solution but `tito build --rpm --tag="foo-1~beta1-1"` works. I didn't test anything else.


FrostyX:
Hello @Blaimi,
thank you for the PR and sorry for the late reply.

Replacing `~` with `@` suggested also @praiskup in #377, so I think we can go with it even though it is not the cleanest. But when I am trying this code, I still get errors when running

    tito tag --use-version=2~dev1

FrostyX:
*bump* @Blaimi,
can I do something to help you with this PR?

Blaimi:
I need some more time to better get into the code. Since we are creating the tags and the `.tito/packages`-file with our ci instead of `tito tag` this PR fixes our problem with `tito build`. We absolutely don't want to monkeypatch tito everytime we use it, so this PR will absolutely receive some more attention by us ;)

FrostyX:
Alright :-)
I am glad to hear that.

If you need anything, let me know.

Blaimi:
I fixed the problem with tito tag. the resulting git tag is now with `@` while the version number/tagname in `.tito/packages/foo` still has `~` – which is fine.

I'm not yet sure, if all the `replace`s in `common.py` are really necessary, some of the methods are receiving already replaced names. I'm not having the meta-view on the project yet and I'm not sure if I can find enough time for it in a near time.

What is missing:
* unittests
* automatic version bump (e.g. `1.2.3~beta1 -> 1.2.3~beta2`)
* testing, testing, testing
* cleanup of some uneccessary `replace()` if any
* error-messages or autocorrect if someone uses `tito tag --use-version=2@dev1` (it fails becaus the rpm validator fails with `error: line 2: Illegal char '@' (0x40) in: Version:        2@dev4`)

emilbreiner99:
Hello @FrostyX ,

I would like to know if I can help finish this MR if something is missing.

Blaimi:
> I would like to know if I can help finish this MR if something is missing.

> * unittests
> * automatic version bump (e.g. `1.2.3~beta1 -> 1.2.3~beta2`)
> * testing, testing, testing
> * cleanup of some uneccessary `replace()` if any
> * error-messages or autocorrect if someone uses `tito tag --use-version=2@dev1` (it fails becaus the rpm validator fails with `error: line 2: Illegal char '@' (0x40) in: Version:        2@dev4`)

:-p

“testing” means especially the parts beside `build`. I did not get all the abstractions, i.e. where a tag is converted to a version and vice versa, there seem to be multiple places.

bryce-carson:
```py
    def _validate_options(self):
        if self.options.keep_version and self.options.use_version:
            error_out("Cannot combine --keep-version and --use-version")

        if '@' in self.args.use_version or '@' in self.args.use_release:
            error_out("Cannot use symbol '@' in release or tag. Use '~'.")
```

Would adding the second condition (as shown above) to the `_validate_options` method of the `TagModule` be sufficient to resolve that?

- [ ] error-messages or autocorrect if someone uses `tito tag --use-version=2@dev1` (it fails becaus the rpm validator fails with `error: line 2: Illegal char '@' (0x40) in: Version:        2@dev4`)

 I'm refreshing my Python skills. I normally code in R, and I'm still unfamiliar with using Tito, so I'm cautious about opening a PR.

At ~7000 lines Tito is a good project for me to study.

PR #388, opened 2020-09-23
Fix copy_extra_sources for remote URLs

### LINKED ISSUE #387: copy_extra_sources doesn't handle remote URLs well
Ohai,

in our spec files, we use `SourceX: https://some.example.com/artifact.tar.gz` as sources, but the new `copy_extra_sources` code fails on that:

```
      File "/usr/lib/python3.8/site-packages/tito/builder/main.py", line 217, in srpm
        self.copy_extra_sources()
      File "/usr/lib/python3.8/site-packages/tito/builder/main.py", line 516, in copy_extra_sources
        shutil.copy(src, self.rpmbuild_sourcedir)
      File "/usr/lib64/python3.8/shutil.py", line 415, in copy
        copyfile(src, dst, follow_symlinks=follow_symlinks)
      File "/usr/lib64/python3.8/shutil.py", line 261, in copyfile
        with open(src, 'rb') as fsrc, open(dst, 'wb') as fdst:
    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tito/…'
```

We're using a (derived) `GitAnnexBuilder` so I have a hunch that maybe `NoTgzBuilder` should just have a `pass` as `copy_extra_sources`?

FrostyX:
Fix #387

A build with a remote URL in its `SourceX` will work only with

    %_disable_source_fetch 0

macro defined. It can be done either globally via `~/.rpmmacros` or by
using tito with following parameter

    --rpmbuild-options="--define '%_disable_source_fetch 0'"

IMHO this macro should be set by default but we should also introduce
a switch to turn it off (like e.g. Mock or Copr do). Since it is not
related to fixing #387, I am not doing it in this commit.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -34,7 +34,7 @@ from tito.common import scl_to_rpm_option, get_latest_tagged_version, \
     find_cheetah_template_file, render_cheetah, replace_spec_release, \
     find_spec_like_file, warn_out, get_commit_timestamp, chdir, mkdir_p, \
     find_git_root, info_out, munge_specfile, BUILDCONFIG_SECTION
-from tito.compat import getstatusoutput, getoutput
+from tito.compat import getstatusoutput, getoutput, urlparse
 from tito.exception import RunCommandException
 from tito.exception import TitoException
 from tito.config_object import ConfigObject
@@ -508,6 +508,15 @@ class Builder(ConfigObject, BuilderBase):
             if os.path.isfile(dst_file):
                 debug('Source file "%s" already exists. Skiping.' % dst_file)
                 continue
+
+            parse = urlparse(source)
+            if parse.scheme and parse.netloc:
+                # If macro `%_disable_source_fetch 0` is defined, the file will
+                # be automatically downloaded by rpmbuild to the SOURCES
+                # directory. We can safely skip it here.
+                debug('Source "%s" is not a local file. Skiping.' % source)
+                continue
+
             src = os.path.join(self.rpmbuild_sourcedir, self.tgz_dir, source)
             if os.path.islink(src) and os.path.isabs(src):
                 src = os.path.join(self.start_dir, os.readlink(src))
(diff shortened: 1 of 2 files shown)


FrostyX:
What do you think @evgeni, does this change fix #387 for you?

evgeni:
Thanks @FrostyX, I didn't *test* the change, but the code looks like it would successfully skip my sources!

Not sure what that macro is you're mentioning, never seen it before.

FrostyX:
Let me use the `Source1` from #387 as an example.

    Source1:    https://rubygems.org/gems/concurrent-ruby-1.1.5.gem

I don't have many experiences with ruby packaging, so I might be wrong but my guess is that you somehow need to download this `concurrent-ruby-1.1.5.gem` file and put it into `SOURCES` directory? If you wouldn't use `tito` but only `rpmbuild`, you could either download the file into `~/rpmbuild/SOURCES/` or put `%_disable_source_fetch 0` into your `~/.rpmmacros`. Then all missing sources would be automatically downloaded. When using `tito`, the `SOURCES` directory (and the whole `rpmbuild` directory tree) is not static and is generated in `/tmp` for each build, so I don't see any way to download the `concurrent-ruby-1.1.5.gem` manually and putting it somewhere. That's why I was suggesting to use the macro, so that `tito` automatically downloads it to the temporary `SOURCES` directory. Makes sense?

But then, I might be mistaken about how you actually build your package. Would you mind testing this PR? I want to make sure everything works for you before merging and releasing. It doesn't have to be ASAP, we can leave it here until you have time to do so. It should be as simple as

```
git clone https://github.com/FrostyX/tito.git
git checkout copy-extra-sources-remote-url
cd tito
dnf builddep tito.spec
tito build --rpm --test --install
```

evgeni:
Ah, okay, that's an `rpmbuild` feature!

So in our case, our sources are managed by `git-annex` and the `SourceX` line is just what's used to tell annex where to find the file on the web. So annex takes care of all the fetching for us.

FrostyX:
I see, then you really just needed the `copy_extra_sources` to skip the remote URL and nothing else. I was reproducing the failure with just simple `tito.builder.Builder` so I needed other way to download them.

FrostyX:
Thank you again @evgeni,
I am going to merge the PR now.

If you find out that it still needs some tweaking, please reopen the issue :-)

PR #389, opened 2020-10-01
Do not use absolute path to generate-patches.pl script

jirihnidek:
* When tito is installed using pip, then generate-patches is
  installed into the `~/.local/bin` directory and the absolute
  path cannot work


FrostyX:
Thank you for the patch @jirihnidek.
Makes sense, +1

PR #391, opened 2020-10-06
Make tests working on systems without non-versioned python macros

FrostyX:
See RHBZ 1784688

We are using unversioned python macros, such as `%__python` and
`%python_sitelib` which are not supported in Fedora rawhide
anymore. Instead, we should use `%__python3` and `%python3_sitelib`.


FrostyX:
I fixed the Travis CI tests, leaving it here for a few days if anyone wants to review. I will merge on Friday otherwise.

FrostyX:
I left a few days here if somebody wants to review but IMO this is not a very interesting PR. I am merging.

PR #392, opened 2020-10-21
Ka8zrt local repo

cinnion:
This commit allows for releasing to a local repository where there is no need to create a local copy, update it, and push the changes to the original repository. The createrepo2 script is a script which rebuilds the repository, including noarch architectures as needed. It can be found at: 

https://gist.github.com/cinnion/49b2b633619c64d3437b5b8a0307d386 

An example of the use can be seen in the ka8zrt-new branch.

### DIFF
--- a/src/tito/release/LocalYumRepo.py
+++ b/src/tito/release/LocalYumRepo.py
@@ -1,0 +1,134 @@
+import os.path
+import subprocess
+import shutil
+
+from tito.common import run_command, info_out, error_out, debug
+from tito.release import Releaser
+
+class LocalYumRepoReleaser(Releaser):
+    """
+    A releaser which will build the desired packages, copy them to a local
+    Yum repository, and update the repodata.
+
+    WARNING: This will not work in all situations, depending on the current OS,
+    and other factors.
+    """
+    REQUIRED_CONFIG = ['repo_root_dir', 'builder', 'srpm_disttag' ]
+
+    # Architecture mappings. Do not include the noarch or source pseudo-architectures.
+    arch_map = {
+        'i686' : 'i386',
+        'amd64' : 'x86_64'
+    }
+
+    # Default list of file extensions to copy
+    filetypes = [ 'rpm', 'srpm' ]
+
+    # Default list of compiled architectures supported
+    arches = [ 'i386', 'x86_64' ]
+
+    # Default name of directory into which the compiled packages are to be placed.
+    arch_packages_dir = 'Packages'
+
+    # Default name of the directory for the source repo
+    src_arch_dir    = 'Sources'
+
+    # Default name of the directory into which source RPMs are placed
+    src_packages_dir = 'SPackages'
+
+    def __init__(self, name=None, tag=None, build_dir=None,
+            config=None, user_config=None,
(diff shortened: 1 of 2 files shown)


FrostyX:
Hello @cinnion,
I really like the feature, thank you for the PR.

I have a few notes that would IMHO improve it a little bit. 

At the end of the output, I get

```
...
13:31:00: Renamed x86_64/.repodata/ -> x86_64/repodata/
13:31:00: Memory cleanup
13:31:00: All done
```

All printed paths are relative to the `repo_root_dir` which is fine but someone like me will most certainly not remember to what path the `repo_root_dir` was set. I would suggest putting one more line to the output, e.g.

```
13:31:00: All done
13:31:00: Repository path: /tmp/myrepo/x86_64/
```

Also, I was a bit surprised that I need to configure a `builder`, I would expect the `tito.builder.Builder` to be used by default. IMHO it is because you don't call the parent `__init__` method in your constructor. Please see how other releasers do it (e.g. `RsyncReleaser`).

Additionally, could you please describe your releaser in the [releasers.conf.5.asciidoc](https://github.com/rpm-software-management/tito/blob/master/releasers.conf.5.asciidoc), so it is available through `man releasers.conf`?

And the last thing, which I see as a blocker - Can we please get rid of depending on custom `createrepo2` script? I would like to see it implemented within the `LocalYumRepoReleaser` itself. Do you think it would be possible?

cinnion:
Sorry for the delay in responding... just as I sent the PR work has gotten a bit "busy", so it may take me a bit to respond.

> Hello @cinnion,
> I really like the feature, thank you for the PR.

Quite welcome!

> I have a few notes that would IMHO improve it a little bit.
> 
> At the end of the output, I get
> 
> ```
> ...
> 13:31:00: Renamed x86_64/.repodata/ -> x86_64/repodata/
> 13:31:00: Memory cleanup
> 13:31:00: All done
> ```
> 
> All printed paths are relative to the `repo_root_dir` which is fine but someone like me will most certainly not remember to what path the `repo_root_dir` was set. I would suggest putting one more line to the output, e.g.
> 
> ```
> 13:31:00: All done
> 13:31:00: Repository path: /tmp/myrepo/x86_64/
> ```

Good point! I will see what I can do about that.

> Also, I was a bit surprised that I need to configure a `builder`, I would expect the `tito.builder.Builder` to be used by default. IMHO it is because you don't call the parent `__init__` method in your constructor. Please see how other releasers do it (e.g. `RsyncReleaser`).

I will look at that and see what I can do.

> Additionally, could you please describe your releaser in the [releasers.conf.5.asciidoc](https://github.com/rpm-software-management/tito/blob/master/releasers.conf.5.asciidoc), so it is available through `man releasers.conf`?

Definitely.

> And the last thing, which I see as a blocker - Can we please get rid of depending on custom `createrepo2` script? I would like to see it implemented within the `LocalYumRepoReleaser` itself. Do you think it would be possible?

Yes. This will probably be the bulk of my time, and the next couple of weeks look packed (and perhaps through the end of the year... I may have to move during the middle of the pandemic due to my landlord trying to sell my house, and if I do, I am going to go ahead and move back to another city... thankfully, a telecommute job makes this possible).

Issue #332, opened 2019-02-04
Add option to set build number for --test


alexandrevicenzi:
When using --test option the package name will be something like:

    <package name>-<package version>.git.<commit count>.<last commit hash>.el7eng.noarch.rpm

    foobar-0.0.1-1.git.0.5a86da9.el7eng.noarch.rpm

The problem is that sometimes the commit count is the same when building multiple patches. The count is usually 0 (zero) and the hash is not a number, in this case the new version string can be lower than the old one and the test package won't install because is a downgrade.

It would be nice to have an option to set this parameter. For example, when using Jenkins or other CI this parameter could easily be set to the BUILD NUMBER, this will ensure that the version is always greater than the older one.

I'm not sure if this option fits somewhere, maybe in the `tito.props` file or somewhere else.

So far I wrote a custom builder to do this trick and use CI build number.

PR #394, opened 2020-12-26
Added test_version parameter to cli

abdullahzen:
### Changes made:
* Added test_version parameter to cli when used with build --test to produce the test artifact with a desired numbered version rather than using the commit_count which is always 0.
* Kept the default value for the "test_version" as 0.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -91,6 +91,7 @@ class BuilderBase(object):
         # releasers in their config sections.
         if args and 'test' in args:
             self.test = True
+            self.test_version = self._get_optional_arg(kwargs, 'test_version', 0)
 
         # Location where we do all tito work and store resulting rpms:
         self.rpmbuild_basedir = build_dir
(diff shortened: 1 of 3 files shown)


xsuchy:
-1 from me.

Commit_count is the number of commits since the last tagged version. When your workflow is git-commit/tito-build/rpm -U then commit_count assures that package is always upgradable (every subsequent release is bigger than the previous one). Although this breaks when you `git commit --amend -a`

abdullahzen:
@xsuchy 
Thank you for taking the time. I do agree with you but I was adding a feature as requested in the issue #332 where he mentions a use case for overriding the version.
I do see where I went wrong tho. I should set the default to commit_count and if test_version is passed, it'll override the commit_count. Thank you for clarifying that for me 👍 
Do you believe that would be a good idea to allow both ways and let the test_version only override the count if test_version is set?

src/tito/builder/main.py
```diff
@@ -632,8 +633,9 @@ def _get_display_version(self):
         if self.test:
             # should get latest commit for given directory *NOT* HEAD
             latest_commit = get_latest_commit(".")
-            self.commit_count = get_commit_count(self.build_tag, latest_commit)
-            version = "git-%s.%s" % (self.commit_count, latest_commit[:7])
+            if self.test_version is None:
+                self.test_version = get_commit_count(self.build_tag, latest_commit)
+            version = "git-%s.%s" % (self.test_version, latest_commit[:7])
```
xsuchy, line 638: +1 LGTM now.

abdullahzen: Thank you!

PR #396, opened 2021-01-11
Fix Travis CI

FrostyX:
Feels like all I am doing is fixing the Jenkins CI.

The first commit is not necessary, but it is a thing that was bugging me for ages, so I am finally fixing it. The actual fix is the `a2x` change in the specfile. I believe it isn't our bug but rather an `a2x` one, so I will file a ticket for that.


FrostyX:
There is not really a change to the code, so I think we can safely merge it without a review. I would like to have the CI working again so we can merge also other PRs

PR #406, opened 2021-05-05
150 Add support for repos that use submodules.

areese:
Allow though submodules are a contentious topic, this adds a submodule_aware_builder

This can be used with the following config in tito.props:

```
[buildconfig]
builder = SubmoduleAwareBuilder
```

### DIFF
--- a/src/tito/builder/submodule_aware_builder.py
+++ b/src/tito/builder/submodule_aware_builder.py
@@ -1,0 +1,210 @@
+# Copyright (c) 2008-2011 Red Hat, Inc.
+#
+# This software is licensed to you under the GNU General Public License,
+# version 2 (GPLv2). There is NO WARRANTY for this software, express or
+# implied, including the implied warranties of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. You should have received a copy of GPLv2
+# along with this software; if not, see
+# http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt.
+#
+# Red Hat trademarks are not licensed under GPLv2. No permission is
+# granted to use or replicate Red Hat trademarks that are incorporated
+# in this software or its documentation.
+"""
+Tito builders for a variety of common methods of building sources, srpms,
+and rpms.
+"""
+
+import os
+
+from tito.builder import Builder
+from tito.common import debug, run_command, create_tgz, find_spec_like_file, get_commit_timestamp
+from tito.tar import TarFixer
+
+
+class PushDir(object):
+    """
+    Helper class to change directories and restore the same way pushd/popd work in the shell.
+    Usage:
+    # cwd = 'current'
+    with PushDir("into"):
+        # cwd = 'current/into'
+        do_command # cwd is into.
+    # cwd = 'current'
+    """
+
+    def __init__(self, subdir):
+        self.old_wd = os.getcwd()
+        self.new_wd = None
+        if subdir:
+            self.new_wd = os.path.join(self.old_wd, subdir)
+
+    def __enter__(self):
+        if self.new_wd:
+            self.chdir(self.old_wd, self.new_wd)
+
+    def __exit__(self, type, value, traceback):
+        if self.new_wd:
+            self.chdir(self.new_wd, self.old_wd)
+
+    def chdir(self, old_dir, new_dir):
+        if new_dir:
+            debug('Changing directory from %s to %s' % (old_dir, new_dir))
+            os.chdir(new_dir)
+
+
+class SubmoduleAwareBuilder(Builder):
+    """
+    Submodule aware builder class.
+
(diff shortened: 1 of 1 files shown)


src/tito/builder/submodule_aware_builder.py
```diff
@@ -0,0 +1,210 @@
...
+"""
+
+import os
+
+from tito.builder import Builder
+from tito.common import debug, run_command, create_tgz, find_spec_like_file, get_commit_timestamp
+from tito.tar import TarFixer
+
+
+class PushDir(object):
```
FrostyX, line 25: I think we have this already implemented as `tito.common.chdir`. For example `MeadBuilder` uses it as

    with chdir(self.maven_clone_dir):
        run_command("git checkout %s" % self.git_commit_id)

src/tito/builder/submodule_aware_builder.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    Submodule aware builder class.
+
+    This builder is aware of submodules and will package the contents of the submodules as well.
+    It expectes `git submodule--helper list` to exist and work.
+    """
+    REQUIRED_ARGS = []
+
+    # TODO: drop version
+    def __init__(self, name=None, tag=None, build_dir=None,
```
FrostyX, line 66: If I understand it correctly, this constructor just calls the parent's constructor and does nothing else?
In that case, we can remove it altogether :-)

src/tito/builder/submodule_aware_builder.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        self._create_build_dirs()
+
+        debug("Creating %s from git tag: %s..." % (self.tgz_filename,
+                                                   self.git_commit_id))
+
+        # call our create_tgz, instead of the global one.
+        self.create_tgz(self.git_root, self.tgz_dir, self.git_commit_id,
+                        self.relative_project_dir,
+                        os.path.join(self.rpmbuild_sourcedir, self.tgz_filename))
```
FrostyX, line 110: What if we update the `Builder._setup_sources` method and change 

```
create_tgz(self.git_root, self.tgz_dir, self.git_commit_id,
           self.relative_project_dir,
           os.path.join(self.rpmbuild_sourcedir, self.tgz_filename))
```

to

```
self._create_tgz()
```

and then define it in the `Builder` class like this

```
def _create_tgz(self):
        return create_tgz(self.git_root, self.tgz_dir, self.git_commit_id,
                          self.relative_project_dir,
                          os.path.join(self.rpmbuild_sourcedir, self.tgz_filename))
```

You could then simply override this method instead of the whole `_setup_sources`. Or if you prefer any other approach, feel free, this is just a suggestion.

FrostyX (commented):
Thank you @areese for the PR!

> Allow though submodules are a contentious topic

To say at least :D
but I understand that people use them and therefore submodules-aware builder might be a real need. Thank you for implementing it.

I made a few suggestions to shorten the PR a little, what do you think about them?
Also, would you mind briefly describing the builder in the documentation?
https://github.com/rpm-software-management/tito/blob/master/doc/builders.md

src/tito/builder/submodule_aware_builder.py
areese: Oh interesting, I hadn't seen this.

Lemme try and figure that part out and issue it instead.

src/tito/builder/submodule_aware_builder.py
areese: That would be a lot better, I didn't want to mess with the code that was already there.

If you would like to do the refactor it will be faster.
For me, I'll have to get approval for any changes, and that can take some time.  :)

areese:
I'm going to fix the small things, and add the documentation.
I'm going to work on approval for the refactor, that will likely end up being a separate pr as that can take time.

src/tito/builder/submodule_aware_builder.py
areese: This is perfect, I'll use with chdir

areese:
@FrostyX I've done all but the create_tgz refactor.  I'll start working that through the chain I need to follow, thank you.

src/tito/builder/__init__.py
```diff
@@ -15,4 +15,8 @@
     GitLfsBuilder, \
     MeadBuilder
 
+from tito.builder.submodule_aware_builder import \
+    SubmoduleAwareBuilder
+
 from tito.builder.fetch import FetchBuilder
+
```
FrostyX, line 22: The CI tests fail because they do a PEP8 check and it complains about this empty line. Can you please drop it?

areese: Fixed.  I think I ended up with an extra newline when shipping patch files around

areese:
@FrostyX yep.  I'm going to work on a followup, there process I have to follow and it can take lots of time.

PR #407, opened 2021-05-21
Adding option to fetch sources

schlupov:


FrostyX:
Thank you very much for the PR!
It looks good to me and I wouldn't mind merging it as is.

But I was playing with it a little and found the PR #388, please see its description. Basically, if you use the default `tito.builder.Builder` and call tito with the following parameter 

    --rpmbuild-options="--define '%_disable_source_fetch 0'"

Then it provides the exact same functionality as this PR does. In the #388 I said

> MHO this macro should be set by default but we should also introduce
> a switch to turn it off (like e.g. Mock or Copr do). [...], I am not doing it 
> in this commit.

I still think this would be a good idea. What y'all think?

praiskup:
I did not test it, but if it does the same thing - and it is easier to implement - it sounds good.
What we need though is a configuration option in `.tito.props` (or somewhere).

schlupov:
PTAL. I added a new switch --fetch-sources and option into tito.props

src/tito/cli.py
```diff
@@ -355,6 +358,10 @@ def __init__(self):
                 default='',
                 metavar="COLLECTION", help="Build package for software collection.")
 
+        self.parser.add_option("--fetch-sources", dest='fetch_sources',
+                               action="store_true",
+                               help="Download sources from a predefined url address to the SOURCE folder")
```
praiskup, line 363: s/predefined url address/predefined Source<N> addresses/

schlupov:
Updated that help text. @FrostyX can you please look at the pull request?
@praiskup you can try the code from this PR using the command:
`PYTHONPATH=/path/to/tito/src/ python /path/to/tito/src/tito/cli.py`

praiskup:
I can confirm that it worked during my quick testing!

FrostyX:
Thank you very much @schlupov,
and @praiskup for testing.

PR #408, opened 2021-06-22
Rename and document the tito.props option for fetching sources

FrostyX:
Complements PR #407

Every other `tito.props` config option uses underscores instead of
dashes, so let's preserve this convention.

Also describing it in the manpage.


FrostyX:
What do you think @schlupov and @praiskup?

praiskup:
+1

schlupov:
+1, good point

PR #420, opened 2022-03-22
DistGitReleasers: don't nuke external sources with fetch_sources

praiskup:
Previously only 'tito build --srpm|--rpm' worked well with fetch_sources
where we rely on RPM's (rpmbuild) built-in download mechanism.  It did
not work well with 'tito release' and DistGit releasers, though.

The thing is that we don't use rpmbuild at all while doing DistGit
releases BUT we still want those files to be (a) downloaded to
%_sourcedir and then (b) uploaded to DistGit lookaside cache.  Because
we did not download those files before, every extra source file was
entirely ignored, or even worse, if the file was already uploaded in
the target lookaside cache, the file was removed from sources.

This bitten us during 'copr-backend' releases several times, and starts
being painful in 'resalloc-aws' releases, too.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -34,7 +34,7 @@ from tito.common import scl_to_rpm_option, get_latest_tagged_version, \
     find_cheetah_template_file, render_cheetah, replace_spec_release, \
     find_spec_like_file, warn_out, get_commit_timestamp, chdir, mkdir_p, \
     find_git_root, info_out, munge_specfile, BUILDCONFIG_SECTION
-from tito.compat import getstatusoutput, getoutput, urlparse
+from tito.compat import getstatusoutput, getoutput, urlparse, urlretrieve
 from tito.exception import RunCommandException
 from tito.exception import TitoException
 from tito.config_object import ConfigObject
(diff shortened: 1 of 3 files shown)


FrostyX:
The CI failure is caused by

```
  File "setup.py", line 22, in <module>
    with open("README.md", "r", encoding="utf-8") as fh:
TypeError: 'encoding' is an invalid keyword argument for this function
```

which is not caused by this PR. Happened to me also in another PR.

FrostyX:
LGTM, thank you for the PR.

src/tito/builder/main.py
```diff
@@ -79,7 +79,12 @@ def __init__(self, name=None, build_dir=None,
 
         self.quiet = self._get_optional_arg(kwargs, 'quiet', False)
         self.verbose = self._get_optional_arg(kwargs, 'verbose', False)
-        self.fetch_sources = self._get_optional_arg(kwargs, 'fetch_sources', False)
+
+        # The default for the builder instantiated from Releasers
+        config_fetch_default = False
+        if self.config.has_option(BUILDCONFIG_SECTION, "fetch_sources"):
+            config_fetch_default = self.config.get(BUILDCONFIG_SECTION, "fetch_sources")
+        self.fetch_sources = self._get_optional_arg(kwargs, 'fetch_sources', config_fetch_default)
```
FrostyX, line 87: Not your fault in the slightest, just pointing out something that annoys me - Ten minutes ago I wrote something similar in another PR in `MockBuilder`. It's awful, we can't parse the config and deal with argument defaults in a thousand different places.

We need a to unify it and properly set defaults from the config file for all recognized arguments.

praiskup: Yes, that's a reasonable TODO item.  I didn't want to change too many things in this PR, but I was tempted to do this right away :-)

FrostyX: Me too but I was scared to do it without thorough testing :-)

PR #423, opened 2022-06-02
Remove '-.*' from the end of git-annex version strings.

pcreech:
RPMs provided by git-annex have a '-{hash}' added to the version, breaking version detection.

This was the simplest way I could come up with to get around this issue, based on what was already implemented.

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -1045,7 +1045,7 @@ def compare_version(version1, version2):
     zero when equal and positive when version1 > version2.
     """
     def normalize(v):
-        return [int(x) for x in re.sub(r'(\.0+)*$', '', v).split(".")]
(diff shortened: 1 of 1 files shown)


pcreech:
The git-annex rpm repo I'm using is:

```
[git-annex]
name=git-annex repository
baseurl=https://downloads.kitenet.net/git-annex/linux/current/rpms/
enabled=0
gpgcheck=0
```

FrostyX:
Hello @pcreech,
sorry for the late reply.

I like the PR, and I would keep the change where it is. Can you please add these two asserts into `common_tests.py:test_compare_versions`?

```python
self.assertEquals(0, compare_version("1.0-abc", "1.0-xyz"))
self.assertTrue(compare_version("1.1-abc", "1.0-xyz"))
```

PR #425, opened 2022-08-23
Revert 45d431ad149cb33e2462a990c4c4f29e6bb2bb7e

nikosmoum:
- The original change made the assumption that the git_branches[0]
  is the actual branch name that we are on, but that was incorrect.


FrostyX:
Hello @nikosmoum,
you are the author of the original commit, so if you want to revert it, it's your call :-)

But the original commit was trying to achieve something - aren't we going to revert to some previous issue?

nikosmoum:
@FrostyX To me it seemed the original commit was trying to add functionality instead of fixing an issue, so I don't think it's a problem to revert

FrostyX:
Okay. I am not using Mead and the original commit was yours, so we do what you say :-)

PR #431, opened 2022-11-30
use spdx license

xsuchy:
https://fedoraproject.org/wiki/Changes/SPDX_Licenses_Phase_1

### DIFF
--- a/tito.spec
+++ b/tito.spec
@@ -20,7 +20,7 @@ Version: 0.6.22
 Release: 1%{?dist}
 Summary: A tool for managing rpm based git projects
 
-License: GPLv2
+License: GPL-2.0-only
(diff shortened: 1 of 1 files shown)


tito.spec
```diff
@@ -20,7 +20,7 @@ Version: 0.6.22
 Release: 1%{?dist}
 Summary: A tool for managing rpm based git projects
 
-License: GPLv2
+License: GPL-2.0-only
```
FrostyX, line 23: Funny thing, the `setup.py` says `license='GPLv2+'`.
Should we change it? Or should we do `GPL-2.0-or-later` here?

xsuchy: I think setup.py is newer. I would use the older one as a source of truth.

FrostyX (approved):

PR #434, opened 2023-01-04
Don't upload patches to the lookaside cache

FrostyX:
With @xsuchy we discovered that when doing `tito release`, into Fedora DistGit, then .patch files are being uploaded to the lookaside cache. That is not expected behavior, we should commit them to git only.

### DIFF
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -327,13 +327,16 @@ class FedoraGitReleaser(Releaser):
         Upload any tarballs to the lookaside directory. (if necessary)
         Uses the "fedpkg new-sources" command.
         """
-        if not self.builder.sources:
+        sources = [x for x in self.builder.sources
+                   if not x.endswith(self.copy_extensions)]
+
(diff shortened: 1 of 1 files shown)


xsuchy (approved):
+1

src/tito/release/distgit.py
```diff
@@ -343,7 +346,7 @@ def _git_upload_sources(self, project_checkout):
         output = run_command(cmd)
         debug(output)
         debug("Removing write-only permission on:")
-        for filename in self.builder.sources:
+        for filename in sources:
             run_command("chmod u+w %s" % filename)
```
praiskup, line 350: off-topic: is the "removing write-only permission" a correct debug entry?  Seems like we are actually adding write permission to the user.

praiskup (approved):

src/tito/release/distgit.py
FrostyX: You are right, fixed.

PR #439, opened 2023-01-18
automation: some useful pre-commit workflows

praiskup:
- enable differential PyLint checker
- enable differential ShellCheck (there's a bash script in this repo)
- enable Fedora test matrix


praiskup:
I'm not sure about the tex test.  Could be useful, but if we can't run it first ... :-/

FrostyX:
Thank you :-)

PR #461, opened 2023-06-29
Fix UpstreamBuilder deprecation warning

FrostyX:
I started seeing this deprecation warning in `tito build ...` output.

    RPM build warnings:
        %patchN is deprecated (1 usages found), use %patch N (or %patch -P N)


xsuchy (approved):
It would by nice to use python-specfile to handle this insertation, but yeah, that is different issue.

PR #474, opened 2023-12-28
Allow tito build without .tito directory

### LINKED ISSUE #472: Make `tito build --srpm` work without any .tito/ configuration
To make `tito` the default SCM method in Copr, it would be very convenient
to use it with any random git repository holding just a spec file and
corresponding sources to `git archive`.  I.e. avoid doing the `tito init`
step and `git push`.  Is this, by chance, possible?

FrostyX:
Fix #472

### DIFF
--- a/src/tito/cli.py
+++ b/src/tito/cli.py
@@ -90,8 +90,9 @@ class ConfigLoader(object):
         rel_eng_dir = os.path.join(find_git_root(), tito_config_dir())
         filename = os.path.join(rel_eng_dir, TITO_PROPS)
         if not os.path.exists(filename):
-            error_out("Unable to locate branch configuration: %s"
-                "\nPlease run 'tito init'" % filename)
+            error_out("Unable to locate branch configuration: %s\n"
+                      "Please run 'tito init' or use '--without-init' parameter"
+                      % filename)
 
         # Load the global config. Later, when we know what tag/package we're
         # building, we may also load that and potentially override some global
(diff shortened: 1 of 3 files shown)


src/tito/cli.py
```diff
@@ -267,8 +267,24 @@ def main(self, argv):
...
+            "changelog_do_not_remove_cherrypick": 0,
+            "changelog_format": "%s (%ae)",
+        })
+        return config
+
     def load_config(self, package_name, build_dir, tag):
-        self.config = ConfigLoader(package_name, build_dir, tag).load()
+        self.config = self.initial_config
+        if not self.options.without_init:
+            self.config = ConfigLoader(package_name, build_dir, tag).load()
```
xsuchy, line 287: I think this should be:
```
        if self.options.without_init:
            self.config = self.initial_config
        else:
            self.config = ConfigLoader(package_name, build_dir, tag).load()
```

And the error message of ConfigLoader() should be altered and when it prints:
```
ERROR: Unable to locate branch configuration: /home/msuchy/fedora/git/gnome-logs/rel-eng/tito.props
Please run 'tito init'
```
We should add a line:
```
or run tito --without-init
```

FrostyX: Good points, thank you. Updated.

xsuchy (approved):
+1

praiskup:
Why we can't simply do this even without an explicit `--without-init`?

FrostyX:
> Why we can't simply do this even without an explicit --without-init?

I don't have a strong opinion on this. My reasoning was that running Tito without init, albeit very useful in some cases, it is still a non-typical usage. And I wanted to communicate that to users.

But there is no hard requirement for this, we can make this behavior implicit if we want to.

xsuchy:
> Why we can't simply do this even without an explicit `--without-init`?

Because this way user has to explicitly admit that he is operating with default setting of Tito project. And not with the setting of the git project. 
If it is author, then it encourage him to create configuration (and study what he can do with the configuration).

praiskup:
I don't know. If it is possible, and we can issue a warning to "push the user" toward a useful configuration (while the option could silence the warnings), the explicit option only complicates integrating this feature into higher level tools like Copr (special option needs to be implemented just for this).

PR #475, opened 2023-12-28
Release to all active Fedora and EPEL branches

### LINKED ISSUE #430: enhance FedoraGitReleaser to build to all active Fedora branches
I use this `releasers.conf`:
```
[fedora-git]
releaser = tito.release.FedoraGitReleaser
branches = rawhide f37 f36 f35
```
And it is pain to update it every 3 months because of new releases and new branching.

It would be awesome if I can use some magic to mark all active branches. E.g.,

```
branches = rawhide **fedora-active** **epel-active**
```
The list of active branches can be fetched from https://pdc.fedoraproject.org/ which has API.

FrostyX:
Fix #430

### DIFF
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -11,6 +11,7 @@
 # granted to use or replicate Red Hat trademarks that are incorporated
 # in this software or its documentation.
 import datetime
+import requests
 import os.path
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


FrostyX:
I should add tests and also make sure this doesn't break non-Fedora releasers.

xsuchy:
Hmm, I would welcome if there are aliases like Packit have: https://packit.dev/docs/configuration#aliases 
I thought that we could steal this code, but they have it hardcoded https://github.com/packit/packit/blob/main/packit/config/aliases.py#L14

FrostyX:
I am closing this in favor #476

PR #476, opened 2023-12-30
Support branch aliases in releasers

### LINKED ISSUE #430: enhance FedoraGitReleaser to build to all active Fedora branches

FrostyX:
Fix #430

### DIFF
--- a/src/tito/release/distgit.py
+++ b/src/tito/release/distgit.py
@@ -16,6 +16,13 @@ import subprocess
 import sys
 import tempfile
 
+try:
+    # Optional dependency available on Fedora and EPEL9+
+    # Without it, branch aliases in releasers.conf won't work.
+    from fedora_distro_aliases import get_distro_aliases
+except ImportError:
+    get_distro_aliases = None
+
 from tito.common import (
     run_command,
     debug,
@@ -56,8 +63,7 @@ class FedoraGitReleaser(Releaser):
         else:
             self.cli_tool = "fedpkg"
 
-        self.git_branches = \
-            self.releaser_config.get(self.target, "branches").split(" ")
+        self.git_branches = self._branches()
 
         # check .tito/releasers.conf
         if self.releaser_config.has_option(self.target, "remote_git_name"):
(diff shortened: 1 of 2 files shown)


FrostyX:
This obsoletes PR #475.

What do you think @xsuchy?

@lachmanfrantisek, @lbarcziova, there is a hardcoded version of this in the [Packit code](https://github.com/packit/packit/blob/main/packit/config/aliases.py#L14). @xsuchy suggested creating a separate python package for resolving the aliases. We could use this implementation, what do you think?

src/tito/release/aliases.py
```diff
@@ -0,0 +1,108 @@
...
+from tito.exception import TitoException
+
+
+def pdc_active_products():
+    """
+    Fetch the list of active products from PDC
+    https://pdc.fedoraproject.org/rest_api/v1/product-versions/
+    """
+    pdc = "https://pdc.fedoraproject.org"
+    url = "{0}/rest_api/v1/product-versions/?active=true".format(pdc)
```
xsuchy, line 22: Is branched, but not released active? How we can differentiate the branched but not released one?

xsuchy:
It can be better (like implement rest of the aliases) but this is HUGE improvement. +1 for merging and improve later in next PR :)

src/tito/release/aliases.py
praiskup: I second this question.  From the "releaser" perspective, we want to ensure we publish also into the freshly branched fedora version (which I'm not sure is provided by PDC).

lbarcziova:
@FrostyX the linked variable is used only as a feedback value, otherwise we use Bodhi for resolving the aliases [here](https://github.com/packit/packit/blob/d1f7091ba2571455fade8d452d10740bead566fe/packit/config/aliases.py#L210). But having this code in a separate package for the benefit of reusing it by multiple projects makes sense to me!

FrostyX:
> otherwise we use Bodhi for resolving the aliases here.

Thank you, I will take a look if it is applicable for Tito

> having this code in a separate package for the benefit of reusing it by multiple projects makes sense to me!

Perfect! I will ping you so we can figure out the details (e.g. where it should live) and I will try to kick off the project.

FrostyX:
@lbarcziova I took your `get_aliases` and modified it a bit. It doesn't return e.g.

```python
{
    'fedora-all': ['fedora-38', 'fedora-39', 'fedora-rawhide']
    ...
}
```

but rather

```python
{
    'fedora-all': [
        Distro({'name': 'F38', 'long_name': 'Fedora 38', 'version': '38', ...}),
        Distro({'name': 'F39', 'long_name': 'Fedora 39', 'version': '39', ...}),
        Distro({'name': 'Rawhide', 'long_name': 'Fedora Rawhide', 'version': 'rawhide', ...})
    ...
    ]
}
```

The reason is that in Tito I want to use something like

```python
aliases = get_distro_aliases()
branches = [x.branch for x in aliases["fedora-all"]
```

and in Packit you will want

```python
aliases = get_distro_aliases()
branches = [x.namever for x in aliases["fedora-all"]
```

Also, you will maybe be able to replace (some parts of) your `get_branches` function.

What do you think of the code now? If it was moved to a separate package outside of Tito, would you use it?

lachmanfrantisek:
@FrostyX I like the object-like approach. It would be really handy for us as well -- sometimes, we need the `namever` (for chroots), and sometimes branches (e.g. when working with dit-git).

FrostyX:
But it is ready for a review. I am only planning to squash the commits before merging.

praiskup (approved):
LGTM, thank you.

praiskup:
First package release using Tito with this patch:
https://github.com/fedora-copr/resalloc-ibm-cloud/commit/5ceece9613192befcb97ad3148d099bd99df42d0

PR #477, opened 2024-01-04
unify README and index.md

xsuchy:
README is in better state - I used that one


xsuchy:
The diff seems to be big, but I just did:

```
git rm doc/index.md
git mv README.md doc/index.md
ln -s doc/index.md README.md
```

FrostyX:
> In the specfile we have: %doc doc/*

So the index is currently being installed. After this PR it will disappear. Maybe it would be worth it to add `%doc README.md`?

FrostyX (approved):
Much better, thank you for the PR.

PR #478, opened 2024-01-04
chore: add new issues into the CPT's Kanban

praiskup:
This commit requires specifying the ADD_TO_PROJECT_PAT secret - already done.

### DIFF
--- a/.github/workflows/new-issues-to-board.yml
+++ b/.github/workflows/new-issues-to-board.yml
@@ -1,0 +1,15 @@
+---
+name: Move new issues to issues review
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  move-new-issue:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


FrostyX:
I can merge this, no problem. But do we really want to have all Tito issues in our tracker @praiskup? There are a lot of them.

xsuchy:
I agree with @FrostyX - we do not want to work on every issue of Tito. Just the important ones, that we can add to our Kanban manually.

praiskup:
This only adds new issues, I think.  WRT capacity planning, I think all our issues deserve the triage. Also, it's not likely other team members are likely to reply to new issues if they are not in Kanban...  but... if this seems wrong, let's discuss it on the meeting on Monday?

BTW. I already added resalloc-related packages in `praiskup/` namespace, and added a [mock PR](https://github.com/rpm-software-management/mock/pull/1280)

praiskup:
I'm the only one who thinks would be a profitable change, so closing.

Issue #428, opened 2022-10-20
Bodhi updates via releaser while doing `tito release`?


praiskup:
Can we make this possible?  Bodhi is the only manual step we are doing while releasing packages into Fedora.
It is quite easily to submit bodhi using `fedpkg update` (one command => one update template => many packages in multiple Bodhi updates).  Seems like something tito could automatize.

praiskup:
@xsuchy fyi

xsuchy:
There is python bindings in bodhi-client package.

FrostyX:
I really like the idea

PR #479, opened 2024-01-04
document how to automate bodhi updates

xsuchy:
Fixes: #428


FrostyX (approved):
Thank you for the PR!
Also, the process is much easier, then I expected.

Issue #480, opened 2024-01-05
Please assure upgrade path to tito.rpm from @rpm-software-management/tito Copr


praiskup:
The copr repository is mostly useless now for a random pre-release tester, which is a pity (for example, I'm personally running "vanilla" `mock` from the main branch built in Copr, ditto `csdiff`, and even `tar` every day - but I can't with tito).

Assuming #460 is not going to happen soon, I'd propose a similar process to what we have in Mock -- [post-release](https://rpm-software-management.github.io/mock/Releasing-Mock) version bump like:
```patch
-Version: 5.4
+Version: 5.4.post1
```
I don't like saying this, because it is a suboptimal low-hanging fruit (most people do not expect this to happen when doing `dnf copr enable`), but alternatively, we could at least set `priority=` for the Copr repo.

PR #483, opened 2024-01-10
Bump the ENVRA built directly from git master branch

praiskup:
Fixes: #480

### DIFF
--- a/doc/maintenance.md
+++ b/doc/maintenance.md
@@ -10,7 +10,8 @@ If you are releasing a new version of Tito, please follow these steps:
    Copr `tito release copr --test`
 3. Make sure that `[fedora]` releaser in `.tito/releasers.conf` contains all
    currently supported Fedora and Epel branches.
-4. Tag a new version `tito tag` and follow its instructions
+4. Tag a new version `tito tag --use-version=` (drop the .postX suffix) and
+   follow its instructions
 5. Go to the [GitHub releases page][releases] and
    - write a propper release notes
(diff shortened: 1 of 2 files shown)


praiskup:
This would allow us to consume packit-built packages in `sudo dnf copr enable @rpm-software-management/tito`.

FrostyX:
Thank you for the PR @praiskup,
can we simply do `tito tag` for tagging new versions or do we have to `tito tag --use-version=...` to get rid of the suffix?

praiskup:
Yes, you have to do `tito tag --use-version`, otherwise:
```
-Version: 0.6.24.post1
+Version: 0.6.25.post1
```

praiskup:
Good point, updated.

FrostyX:
> Good point, updated.

That's exactly what I meant. Thank you :-)

praiskup:
Uh oh? :-)

```
$ sudo dnf -y copr enable @rpm-software-management/tito
$ sudo dnf update tito
Updating and loading repositories:
 Copr repo for tito owned by @rpm-software-management                                                                                                                                               100% |   8.6 KiB/s |   4.0 KiB |  00m00s
Repositories loaded.
Nothing to do.
$ rpm -q tito
tito-0.6.24-2.fc39.noarch
```

praiskup:
Hm we produce package named `tito-tito` :D seems like tito versioning is not used at all. Lemme submit another patch then.

Issue #486, opened 2024-01-16
Can't do `tito tag` with tito from git-master


praiskup:
```
Creating output directory: /tmp/tito
Traceback (most recent call last):
  File "/usr/bin/tito", line 33, in <module>
    sys.exit(load_entry_point('tito==0.6.24', 'console_scripts', 'tito')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 944, in main
    CLI().main(sys.argv[1:])
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 210, in main
    return module.main(argv)
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 709, in main
    self.load_config(package_name, build_dir, None)
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 307, in load_config
    self._check_config_and_load(loader)
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 288, in _check_config_and_load
    self.config = loader.load()
                  ^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 76, in load
    self.config = self._read_config()
                  ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/tito/cli.py", line 101, in _read_config
    config.read(filename)
                ^^^^^^^^
NameError: name 'filename' is not defined
```

PR #487, opened 2024-01-16
Fix traceback related to undefined variable

praiskup:
Reverts part of 68d8b219dbea9271d17f24e68cc2b43906255e38

### DIFF
--- a/src/tito/cli.py
+++ b/src/tito/cli.py
@@ -97,6 +97,14 @@ class ConfigLoader(object):
         # Load the global config. Later, when we know what tag/package we're
         # building, we may also load that and potentially override some global
         # settings.
+        # List of filepaths to config files we'll be loading:
+        rel_eng_dir = os.path.join(find_git_root(), tito_config_dir())
+        filename = os.path.join(rel_eng_dir, TITO_PROPS)
+        if not os.path.exists(filename):
+            error_out("Unable to locate branch configuration: %s\n"
+                      "Please run 'tito init' or use '--without-init' parameter"
+                      % filename)
+
         config = RawConfigParser()
         config.read(filename)
 


src/tito/cli.py
```diff
@@ -97,6 +97,14 @@ def _read_config(self):
...
+        # List of filepaths to config files we'll be loading:
+        rel_eng_dir = os.path.join(find_git_root(), tito_config_dir())
+        filename = os.path.join(rel_eng_dir, TITO_PROPS)
+        if not os.path.exists(filename):
+            error_out("Unable to locate branch configuration: %s\n"
+                      "Please run 'tito init' or use '--without-init' parameter"
+                      % filename)
+
         config = RawConfigParser()
         config.read(filename)
```
praiskup, line 109: The `filename` here was undefined.

praiskup:
Fixes #486

src/tito/cli.py
```diff
@@ -97,6 +97,14 @@ def _read_config(self):
         # Load the global config. Later, when we know what tag/package we're
         # building, we may also load that and potentially override some global
         # settings.
+        # List of filepaths to config files we'll be loading:
+        rel_eng_dir = os.path.join(find_git_root(), tito_config_dir())
+        filename = os.path.join(rel_eng_dir, TITO_PROPS)
+        if not os.path.exists(filename):
+            error_out("Unable to locate branch configuration: %s\n"
+                      "Please run 'tito init' or use '--without-init' parameter"
```
praiskup, line 105: This comment doesn't feel right :-) but it fixes the bug...  so perhaps we don't need to check for the file existence?

FrostyX:
I'd propose a different fix. Maybe I am missing something though. Do you think this will work?
(patch is against `master`)

```diff
diff --git a/src/tito/cli.py b/src/tito/cli.py
index 762f49b..3d3a520 100644
--- a/src/tito/cli.py
+++ b/src/tito/cli.py
@@ -98,7 +98,7 @@ def _read_config(self):
         # building, we may also load that and potentially override some global
         # settings.
         config = RawConfigParser()
-        config.read(filename)
+        config.read(self.tito_props_path)

         self._check_legacy_globalconfig(config)
         return config
```

FrostyX:
I think we can close this in favor of PR #489

praiskup:
Agreed.

PR #493, opened 2024-02-21
<gitroot>/tito wrapper

praiskup:
This makes executing of Tito-from-Git a bit more convenient.

### DIFF
--- a/tito
+++ b/tito
@@ -1,0 +1,14 @@
+#! /bin/sh
+
+gitrootdir=$(readlink -f "$(dirname "$0")")
+run_python=${PYTHON-python3}
+
+echo >&2 "\
+Warning: Running Tito from a git repository is not supported.
+If you are not a developer working on Tito code, please install it properly
+https://github.com/rpm-software-management/tito/blob/master/doc/index.md#install
+"
+
+export PYTHONPATH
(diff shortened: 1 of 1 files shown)


FrostyX:
I've always been running Tito from git using e.g.

```bash
PYTHONPATH=/home/jkadlcik/git/tito/src/ python3 /home/jkadlcik/git/tito/src/tito/cli.py build --rpm
```

so personally, I don't have a need for the wrapper. If you find it useful, we can merge though. But in that case, please add some docstring at the top of the script discouraging people from using it unless they are working on Tito code.

FrostyX:
> There already is an ugly stderr output, do you prefer doc, the ugly output, or both? (I prefer just one of them)

I don't have a strong preference, it can be either stderr output, docstring, or both. The problem I have with the stderr as it is right now is that the message is cryptic and it doesn't provide any explanation. I would prefer something like

```
Warning: Running Tito from a git repository is not supported. 
If you are not a developer working on Tito code, please install it properly 
https://github.com/rpm-software-management/tito/blob/master/doc/index.md#install
```

praiskup:
> the message is cryptic and it doesn't provide any explanation

Good point, PTAL

FrostyX (approved):

FrostyX:
Thank you for the changes.

PR #495, opened 2024-02-21
Tito report --nvr

praiskup:


praiskup:
The output is expected, though I'm not sure about the `Creating output directory: /tmp/tito`.  I'll try to take another look once the time permits me to.

praiskup:
I retook a quick look today, but #502 IMO proves this solution is
sub-optimal.  We are duplicating the `munge_specfile` logic
together with get_spec_version_and_release.  And we should do things in
one place.

Also, ideally, things like this should work fine on a read-only filesystem (so
no temporary files or dirs are needed).  I think we should refactor
`munge_specfile` so it doesn't decide the Release/Version itself, but it accepts
these values from the caller (pre-parsed & read-only).  Then, the NVR is "known"
before we even start munging.

The builder.display_version has a tricky value.  That doesn't match anything
else, so it is not reusable (`tito-git-<sha-ref>` basically).  We should make it
match the overall src.rpm pattern, like `tito-<Version>-<sha-ref>`.  Then we
could use it on many other places (like `report --nvr`).

Then, I tend to think we should have `tito report nvr` sub-command  (instead of
--nvr), because we need to add, e.g., the `--test` option which doesn't
logically match with the other `report` options).

Issue #460, opened 2023-06-21
The `tito build --{s,}rpm --test` should have higher NVR than the system wide package


praiskup:
The typical problem I have:

1. Package `Foo` with `Version: 1.1` and `Release: 1` is tagged to
   upstream git, and released (to Fedora)

2. Then `tito build --test ...` works fine because the generated release
   looks like `1.git.3548.f7f3934`.  This is fine, because:

   ```
   $ rpmdev-vercmp 1.1-1.git.3548.f7f3934 1.1-1
   1.1-1.git.3548.f7f3934 > 1.1-1
   ```

   IOW, the package with `--test` has a higher NVR than the one in Fedora.

4. After some time, the Fedora package gets an asynchronous update by
   Fedora community, bumping `Release: 2` (rebuild for GCC bump, e.g.)

5. The `--test` package still provides the same NVR with
   `1.1-1.git.3548.f7f3934` which is though:

   ```
   $ rpmdev-vercmp 1.1-1.git.3548.f7f3934 1.1-2
   1.1-1.git.3548.f7f3934 < 1.1-2
   ```

If the `--test` package is e.g. built n Copr, or locally, it doesn't easily update
the one installed on the system which complicates testing.

I'm not saying now how to resolve this, or whether we should do it by
default.  But it would be nice to have a way to provide a higher NVR than
the one in Fedora somehow.  E.g. by bumping the Version?  Bumping Release
to `666`?  I don't know.

praiskup:
FWIW, tito project itself has the problem now:
```
$ sudo dnf copr enable @rpm-software-management/tito
$ sudo dnf update tito
...nothing happens...
```
the version in Copr is 0.6.24-1.fc39 compared to system default 0.6.24-2.fc39

FrostyX:
Thinking out loud ... so, can we simply automatize the same thing that #480 did? When using `--test`, tito would add the `.post1` suffix for the version? The suffix could be optional (opt-out probably) and configurable.

praiskup:
For me, this is hard to predict. Can we simply try that? Seems like a nice experiment.

praiskup:
Just tried to chain build the client Copr packages from git-head:

    mock --chain \
    /tmp/tito/python-copr-1.132-1.git.75.658230d.fc41.src.rpm \
    /tmp/tito/copr-cli-1.112-1.git.78.658230d.fc41.src.rpm

But for the `copr-cli` package it always installs:

[...]

PR #502, opened 2024-09-27
New config option `buildconfig.test_version_suffix`

praiskup:
If configured like

    $ cat .tito/tito.props
    [buildconfig]
    test_version_suffix = .tito.git

The `tito build --test ...` NEVRA is always higher than the NEVRA of released RPMs.

Fixes: #460

### DIFF
--- a/src/tito/common.py
+++ b/src/tito/common.py
@@ -585,7 +585,8 @@ def replace_spec_release(file_name, release):
             print(line.rstrip('\n'))
 
 
-def munge_specfile(spec_file, commit_id, commit_count, fullname=None, tgz_filename=None):
+def munge_specfile(spec_file, commit_id, commit_count, fullname=None,
+                   tgz_filename=None, version_suffix=None):
     # If making a test rpm we need to get a little crazy with the spec
     # file we're building off. (Note we are modifying a temp copy of the
     # spec) Swap out the actual release for one that includes the git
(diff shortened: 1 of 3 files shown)


src/tito/common.py
```diff
@@ -604,6 +605,12 @@ def munge_specfile(spec_file, commit_id, commit_count, fullname=None, tgz_filena
             ))
             continue
 
+        if version_suffix:
+            m = re.match(r'^(\s*Version:\s*)(.+?)\s*$', line)
+            if m:
+                print(f"{m.group(1)}{m.group(2)}{version_suffix}")
+                continue
```
FrostyX, line 612: I don't know, do we still care about EPEL7? Anyway, it's easy not to break the backward compatibility here.

```
Compiling /builddir/build/BUILDROOT/tito-0.6.26-1.20240927180525922719.pr502.2.g9069fc1.el7_9.x86_64/usr/lib/python2.7/site-packages/tito/common.py ...
  File "/usr/lib/python2.7/site-packages/tito/common.py", line 611
    print(f"{m.group(1)}{m.group(2)}{version_suffix}")
                                                    ^
SyntaxError: invalid syntax
```

Can you please use the `.format` syntax?

FrostyX (commented):
I didn't test it in action but the code looks good to me. Except for that one EPEL7 thing. Thank you very much for the PR.

src/tito/common.py
praiskup: > I don't know, do we still care about EPEL7? 

EPEL 7 is EOL, so we can't update in Koji -> only chance is to ship in Copr now.  IMVHO good time to forget about Python 2 support...

> Can you please use the .format syntax?

Can do, if you prefer it.

praiskup: We don't even need  `.format()` in this case, used plain `+`.

FrostyX: > IMVHO good time to forget about Python 2 support...

I have no problem if we make that decision, drop EPEL7 from tests and so on. I just didn't want to break it by adding one unnecessary f-string :-)

FrostyX (approved):

PR #508, opened 2024-12-04
depend on git-core instead of git

xsuchy:
https://fedoraproject.org/wiki/Changes/ChangeToGitCore


xsuchy:
I belive that the check failure is not relevant to the change in this PR.

PR #509, opened 2024-12-05
tox: drop Python 3.7 add 3.9

praiskup:
No longer provided in the testing environment.

PR #510, opened 2024-12-05
Trying to fix the CI

FrostyX:


tox.ini
```diff
@@ -12,7 +12,7 @@ max-line-length=120
 
 [tox]
 # sync with /.github/workflows/fedora-tox.yml
-envlist = py{36,37,311,312,313}
```
praiskup, line 15: I'd personally keep 3.6/3.9/3.12 as EPELs use those, see #509

.packit.yml
```diff
@@ -19,10 +19,10 @@ jobs:
         - fedora-all
         - epel-7-x86_64
         - epel-8-x86_64
-
-  - <<: *copr
-    trigger: commit
-    metadata:
+        - epel-9-x86_64
       owner: "@rpm-software-management"
```
praiskup, line 23: Seems like you are moving the pull-request builds into the production project, not sure?

FrostyX: Oh, you are right, that's wrong. But we need to specify the same targets for the `trigger: commit` though.

FrostyX: Because this was submitted for the last commit in the `main` branch https://github.com/rpm-software-management/tito/runs/33928767738 and shouldn't have been.

praiskup (approved):
LGTM, thank you!

I'm just not sure this deserves four commits :-) but up to you.

PR #516, opened 2026-02-16
Fix version checking with setuptools 82

### LINKED ISSUE #514: Setuptools no longer provides module pkg_resource
Setuptools seems to have removed the module in version 82 and higher.

```
$ uv pip list
Package    Version
---------- -------
blessed    1.30.0
rpm        0.4.0
ruff       0.15.0
setuptools 82.0.0
tito       0.6.26
wcwidth    0.6.0

$ tito build --rpm --offline
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/bin/tito", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/cli.py", line 944, in main
    CLI().main(sys.argv[1:])
    ~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/cli.py", line 210, in main
    return module.main(argv)
           ~~~~~~~~~~~^^^^^^
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/cli.py", line 428, in main
    builder = create_builder(package_name, build_tag,
            self.config,
            build_dir, self.user_config, args,
            builder_class=self.options.builder, **kwargs)
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/common.py", line 160, in create_builder
    builder_class = get_class_by_name(config.get("buildconfig",
        "builder"))
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/common.py", line 995, in get_class_by_name
    mod = __import__(module, globals(), locals(), [class_name])
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/builder/__init__.py", line 6, in <module>
    from tito.builder.main import \
    ...<9 lines>...
        MeadBuilder
  File "/home/rovanion/source/nsc/prometheus-exporter/.venv/lib/python3.13/site-packages/tito/builder/main.py", line 24, in <module>
    from pkg_resources import require
ModuleNotFoundError: No module named 'pkg_resources'
```

Downgrading to setuptools 81 "solves" the issue.
```
$ tito build --rpm --offline
...
Successfully built: /tmp/tito/nsc-prometheus-exporter-0.5.3-1.src.rpm

$ uv pip list
Package    Version
---------- -------
blessed    1.30.0
rpm        0.4.0
ruff       0.15.0
setuptools 81.0.0
tito       0.6.26
wcwidth    0.6.0
```

But the real solution seems to be for you to [replace pkg_resources with importlib](https://stackoverflow.com/a/79361906/501017).

See also the warning printed by setuptools 81:

> UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.

FrostyX:
Fix #514

As of today, Fedora Rawhide (F45) only has
`python3-setuptools-80.10.2-1.fc44.noarch`, so at the moment, the issue only happens when installing Tito from PyPI.

### DIFF
--- a/src/tito/builder/main.py
+++ b/src/tito/builder/main.py
@@ -20,9 +20,16 @@ import os
 import sys
 import re
 import shutil
-import rpm
-from pkg_resources import require
 from tempfile import mkdtemp
+import rpm
+
+try:
+    from importlib.metadata import version as metadata_version
+    PYTHON_SETUPTOOLS_82 = True
+except ImportError:
+    # A fallback for EPEL8
+    from pkg_resources import require
+    PYTHON_SETUPTOOLS_82 = False
 
 from tito.common import scl_to_rpm_option, get_latest_tagged_version, \
     find_wrote_in_rpmbuild_output, debug, error_out, run_command_print, \
@@ -406,12 +413,12 @@ class Builder(ConfigObject, BuilderBase):
 
         if self.config.has_section("requirements"):
             if self.config.has_option("requirements", "tito"):
-                if Version(self.config.get("requirements", "tito")) > \
-                        Version(require('tito')[0].version):
+                installed_version = tito_version()
+                if Version(self.config.get("requirements", "tito")) > installed_version:
                     error_out([
                         "tito version %s or later is needed to build this project." %
                         self.config.get("requirements", "tito"),
-                        "Your version: %s" % require('tito')[0].version
+                        "Your version: %s" % installed_version
                     ])
 
         self.display_version = self._get_display_version()
@@ -1494,3 +1501,16 @@ class Yum(Rpm):
 
     def builddep(self, spec):
         return "yum-builddep %s" % spec
+
+
+def tito_version():
+    """
+    A backwards-compatible way to figure out the currently installed and
+    running tito version.
+    """
+    if PYTHON_SETUPTOOLS_82:
+        return Version(metadata_version("tito"))
+
+    # A fallback for EPEL8
+    # pylint: disable=used-before-assignment
+    return Version(require('tito')[0].version)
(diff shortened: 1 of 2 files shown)


FrostyX:
For some reason the tests pass on my machine but not in CI. We should fix that before merging.

src/tito/builder/main.py
```diff
@@ -1494,3 +1501,16 @@ def install(self, packages, **kwargs):
 
     def builddep(self, spec):
         return "yum-builddep %s" % spec
+
+
+def tito_version():
```
praiskup, line 1506: Have you considered something like ~ https://github.com/fedora-copr/copr/pull/4115/changes :)

FrostyX: Good point @praiskup, thank you. I wanted to keep the change as straightforward as possible. Also, the function is used only at one place, so it will be easy to change in the future if we need.

praiskup: > I wanted to keep the change as straightforward as possible.

😁  ehm, I wanted to say that `__version__ = 'foo'` might be a simpler way forward

FrostyX: To be honest, I don't really care. I wanted the most brain-dead change I could think of and move to doing something else :D. If you want to implement the `__version__` solution, I have no problem merging it.

praiskup: Turned out to be easier than I anticipated (we already have `__version__`).  #519

PR #517, opened 2026-02-21
Fixing rhel macro condition and removing blessed from setup.py

ccmiersma:
This is a potential fix to the build failure on EPEL 10. The issue appears to be that the python3-blessed package which is available in EPEL for RHEL9 is not yet available for EPEL 10. The simple fix below was generated with some help from AI. I have, however, verified that it builds on both Centos 9 and 10. If there are any concerns with the code, or if you would prefer to wait until python3-blessed makes it's way into EPEL, feel free to reject this request.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
     packages=find_packages('src'),
     include_package_data=True,
     install_requires=[
-        'blessed'
+        
     ],
 
     # automatically create console scripts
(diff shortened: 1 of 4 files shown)


FrostyX:
Hello @ccmiersma, thank you for the PR.
Can you please rebase the PR from `main`? I pushed some test fixes, so we should see green checkmarks in the CI now.

PR #519, opened 2026-02-23
Avoid using setuptools for getting tito version

praiskup:


FrostyX:
Thank you for the PR @praiskup,
I admit this is way simpler than #516.

I tested it and it works as expected.

PR #520, opened 2026-02-23
Fixing rhel macro condition and removing blessed from setup.py

ccmiersma:
I've rebased this to include recent commits. Hopefully this works for you. This works around the missing python3-blessed in EPEL 10. When that package is there, you may want to remove this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
     packages=find_packages('src'),
     include_package_data=True,
     install_requires=[
-        'blessed'
+        
     ],
 
     # automatically create console scripts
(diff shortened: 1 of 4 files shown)


FrostyX:
Thank you @ccmiersma,
we could take the PR as you are proposing, but it would make our life much, much easier if the `python-blessed` package just got added to EPEL10. I tried to build it for EPEL10 in Copr and there wasn't any problem. So I bumped this ticket - https://bugzilla.redhat.com/show_bug.cgi?id=2348444
and I'd like to wait a while if it gets resolved. Plan B is merging this PR :-)

PR #521, opened 2026-05-13
Add runtime and buildtime dependency on python3-pkg-resources

FrostyX:
See https://src.fedoraproject.org/rpms/tito/pull-request/5


hroncok:
This is likely not needed considering acfea18ec658090a7cd5949c135d6c4e7c78c356 and 51d2ce7a6b7ec34e0573c1155633067d12cd3ed0

FrostyX:
You are right, I should've looked better. Reverted in 3c97227

